PR #9, opened 2016-08-03
checks as tasks

philbrookes:
Checks rebuilt as tasks and output added to the tarball dump.

### DIFF
--- a/analysis.go
+++ b/analysis.go
@@ -70,6 +70,9 @@ type CheckResults struct {
 	Results []Result
 }
 
+// CheckTasks will execute all the CheckTasks returned from the supplied checkFactory against the specified project
+// The results of the checks are combined into a single JSON object and written to the writer return from outFor, any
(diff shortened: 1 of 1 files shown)


analysis.go
```diff
@@ -0,0 +1,196 @@
...
+			Namespace string `json:"namespace"`
+			Name      string `json:"name"`
+		} `json:"involvedObject"`
+		Reason  string `json:"reason"`
+		Message string `json:"message"`
+		Count   int    `json:"count"`
+	} `json:"items"`
+}
+
+type DeploymentConfigs struct {
```
rhcarvalho: These types here should match (be a subset of) an specific API version from OpenShift / Kubernetes.
As I mentioned earlier, I'd strongly suggest these to be taken from their original definitions, unneeded parts trimmed, and commentary linking back to where it was taken from (including version information).

PR #11, opened 2016-08-05
add error when no projects are visible

philbrookes:
an empty error was output when the current user had no visible projects, this outputs a more useful error message.

### DIFF
--- a/tasks.go
+++ b/tasks.go
@@ -1,5 +1,7 @@
 package main
 
+import "errors"
+
 // A Task performs some part of the RHMAP System Dump Tool.
 type Task func() error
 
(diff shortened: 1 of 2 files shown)


tasks.go
```diff
@@ -22,17 +24,21 @@ func GetAllTasks(tarFile *Archive) ([]Task, error) {
 		return nil, err
 	}
 
+	if len(projects) == 0 {
+		return []Task{}, errors.New("no projects visible to the currently logged in user")
```
rhcarvalho:
```go
return nil, errors.New("no projects visible to the currently logged in user")
```

We should not allocate memory for an empty slice.

philbrookes: @rhcarvalho opened a PR to address that, here: https://github.com/feedhenry/fh-system-dump-tool/pull/16

PR #13, opened 2016-08-08
Rhmap 9321

philbrookes:
Queue asynchronous write requests and process them synchronously.

### DIFF
--- a/tgz.go
+++ b/tgz.go
@@ -4,14 +4,18 @@ import (
 	"archive/tar"
 	"bytes"
 	"compress/gzip"
+	"errors"
 	"io"
 	"time"
+	"strings"
 )
 
 type Archive struct {
-	tgzFile   io.Writer
-	tarWriter *tar.Writer
-	gzWriter  *gzip.Writer
+	tgzFile    io.Writer
+	tarWriter  *tar.Writer
+	gzWriter   *gzip.Writer
+	WriteQueue chan ArchiveWriteTask
+	Finished   chan bool
 }
 
 type ArchiveWriter struct {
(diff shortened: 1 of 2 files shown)


jhellar:
LGTM, tested on local vm.

tgz.go
```diff
@@ -20,12 +24,24 @@ type ArchiveWriter struct {
...
+	Content []byte
+}
+
 func (a *ArchiveWriter) Write(p []byte) (n int, err error) {
 	return a.Writer.Write(p)
 }
 
 func (a *ArchiveWriter) Close() error {
-	return a.Archive.AddFileByContent(a.Writer.Bytes(), a.File)
+	if len(strings.Trim(a.File, "	 ")) == 0 {
```
maleck13: why do we need to trim like this. Are you just trying to remove spaces? you can use

strings.TrimSpace()

maleck13: So this can likely become if "" == strings.TrimSpace(a.File)

tgz.go
```diff
@@ -38,8 +54,24 @@ func NewTgz(file io.Writer) (*Archive, error) {
...
+
+	go tgz.listenForWrites()
 
 	return &tgz, nil
+
+}
+
+func (a *Archive) listenForWrites() {
+	for {
+		task, openChannel := <-a.WriteQueue
```
maleck13: whats the second param here?

maleck13: ahh I see it is a boolean for if the channel is open, handy

tgz.go
```diff
@@ -67,6 +103,10 @@ func (a *Archive) AddFileByContent(src []byte, dest string) error {
 }
 
 func (a *Archive) Close() {
+	close(a.WriteQueue)
+
+	<-a.Finished
```
maleck13: so we need this to ensure the final write is complete

philbrookes: it doesn't send the true down this channel until the after the last call to addFileByContent has returned, so I don't see how it could get here before the last write is completed?

tgz.go
philbrookes: ah nice! I'll update this in that case.

philbrookes: I did this way to trim both spaces and tabs, yeah I didn't really like it.

tgz.go
maleck13: Yes that is what I was saying. Just making sure I understood why we were doing this

philbrookes:
something wrong with this PR since the repo was moved to the feedhenry org, going to close and re-open.

rhcarvalho:
@david-martin ^^ FYI this is one of the consequences of private -> public :)
Open PRs cannot be updated, existing forks become orphaned.

david-martin:
@rhcarvalho Thanks for the heads up. Definitely something to avoid in future for any other repos that move org.

PR #14, opened 2016-08-08
Add version number and '--version' flag to output it

philbrookes:


rhcarvalho:
Knowledge sharing, for the `flag` package:

```
$ go doc flag | grep minus
One or two minus signs may be used; they are equivalent.
```

So `-version` and `--version` are equivalent.

rhcarvalho:
We took a different approach for the [openshift-template-tool](https://github.com/feedhenry/openshift-template-tool), where version is a subcommand:

```
$ openshift-template-tool version
Version: dev build 
Commit: dev build
```

PR #15, opened 2016-08-08
Add initial .travis.yml

david-martin:


philbrookes:
:+1: Cool, [that build](https://travis-ci.org/feedhenry/fh-system-dump-tool) looks good!

david-martin:
@philbrookes Thanks. We can improve on the build process as needed

.travis.yml
```diff
@@ -0,0 +1,9 @@
+
+language: go
+go: 
+ - tip
+
+script:
+  - gofmt -d -s $(find . -name '*.go')
+  - test -z "$(gofmt -d -s $(find . -name '*.go'))"
```
rhcarvalho: Why do we run `gofmt` on the same input twice?

.travis.yml
```diff
@@ -0,0 +1,9 @@
+
+language: go
+go: 
+ - tip
```
rhcarvalho: Why don't we use a stable Go release?

david-martin: After chatting with @philbrookes, we agreed to use the latest to pick up on any potential issues building with the latest version

.travis.yml
david-martin: The first run is to show the developer what the formatting problem is, if any.
The second run is to allow picking up on if there was any output and fail the build accordlingly, because `go fmt` always seems to exit with a 0.

.travis.yml
```diff
@@ -0,0 +1,9 @@
...
+
+script:
+  - gofmt -d -s $(find . -name '*.go')
+  - test -z "$(gofmt -d -s $(find . -name '*.go'))"
+  - go test
```
rhcarvalho: Note that this runs tests only for the top-level package, not for all packages in the repository.
Works for now, as there is only a single package in the repo. We need to be careful to change this when we introduce more packages, or they won't be tested.

The original suggestion (RHMAP-9214) was `go test ./...` that recursively tests all packages.

david-martin: @rhcarvalho Good catch. An oversight that we can fix in a future change

.travis.yml
rhcarvalho: I'd have the opposite concern, things might work with a newer version, while not working in a released version, the one we'll officially use to build releases of the tool.

I've seen this happen with 1.4 -> 1.5, 1.6 transition.

E.g. one can build code with `tip` that references new fields that don't exist in the latest release.

OTOH Go has the Go1 compatibility clause that guarantees existing code will keep building with newer versions in the 1.x release family.

rhcarvalho: As a compromise, we can have both: `tip` and `1.6` (the version we're building the official releases -- right?).

david-martin: @vibe13  Can you confirm if we are using Go 1.6 for productised version of this?

david-martin:
@rhcarvalho For any suggestions you have on improving on this, you can create a jira in issues.jboss.org/browse/FH

rhcarvalho:
I'm first commenting here to get an understanding closer to the code. Moving this to a different tool just adds friction to the conversation. Once we know if there's anything that needs to change we can create tickets. Sounds good?

david-martin:
@rhcarvalho Sounds good

.travis.yml
vibe13: @david-martin Yes we are, precisely v1.6.3-1

PR #17, opened 2016-08-10
Rhmap 9321

philbrookes:
re-opened from broken PR: https://github.com/feedhenry/fh-system-dump-tool/pull/13

### DIFF
--- a/main.go
+++ b/main.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"math/rand"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -30,6 +31,9 @@ const (
 
 	// The version of the fh-system-dump-tool
 	version = "0.1.0"
+
+	// The RNG seed
+	randSeed = 422423
 )
 
 var (
@@ -133,6 +137,7 @@ func printError(err error) {
 }
 
 func main() {
+	rand.Seed(randSeed)
 	flag.Parse()
 
 	if *versionCheck {
--- a/tgz.go
+++ b/tgz.go
@@ -4,14 +4,18 @@ import (
 	"archive/tar"
 	"bytes"
 	"compress/gzip"
+	"errors"
 	"io"
+	"strings"
 	"time"
 )
 
 type Archive struct {
-	tgzFile   io.Writer
-	tarWriter *tar.Writer
-	gzWriter  *gzip.Writer
+	tgzFile    io.Writer
+	tarWriter  *tar.Writer
+	gzWriter   *gzip.Writer
+	WriteQueue chan ArchiveWriteTask
+	Finished   chan bool
 }
 
 type ArchiveWriter struct {
@@ -20,12 +24,24 @@ type ArchiveWriter struct {
 	Writer  *bytes.Buffer
 }
 
+type ArchiveWriteTask struct {
+	File    string
+	Content []byte
+}
+
 func (a *ArchiveWriter) Write(p []byte) (n int, err error) {
 	return a.Writer.Write(p)
 }
 
 func (a *ArchiveWriter) Close() error {
-	return a.Archive.AddFileByContent(a.Writer.Bytes(), a.File)
+	if "" == strings.TrimSpace(a.File) {
+		return errors.New("File name cannot be empty")
+	}
+	// create a write task from the data in this file
+	writeTask := ArchiveWriteTask{File: strings.Trim(a.File, " "), Content: a.Writer.Bytes()}
+	// and queue it in the archive write queue
+	a.Archive.WriteQueue <- writeTask
+	return nil
 }
 
 func NewTgz(file io.Writer) (*Archive, error) {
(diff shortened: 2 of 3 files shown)


main.go
```diff
@@ -133,6 +137,7 @@ func printError(err error) {
 }
 
 func main() {
+	rand.Seed(randSeed)
```
rhcarvalho:
```go
rand.Seed(time.Now().UnixNano())
```

And then get rid of the constant.

Otherwise every run of the program will generate the very same numbers...

tgz.go
```diff
@@ -20,12 +24,21 @@ type ArchiveWriter struct {
...
 func (a *ArchiveWriter) Write(p []byte) (n int, err error) {
 	return a.Writer.Write(p)
 }
 
 func (a *ArchiveWriter) Close() error {
-	return a.Archive.AddFileByContent(a.Writer.Bytes(), a.File)
+	// create a write task from the data in this file
+	writeTask := ArchiveWriteTask{Name: a.File, Content: a.Writer.Bytes()}
+	// and queue it in the archive write queue
+	a.Archive.WriteQueue <- writeTask
```
rhcarvalho: Suggestion:

```go
a.Archive.WriteQueue <- ArchiveWriteTask{
	Name: a.File,
	Content: a.Writer.Bytes(),
}
```

Now, the responsibilities seem a bit mixed, it is strange that an `ArchiveWriter` needs direct access to a channel/field in an `Archive`.

tgz.go
```diff
@@ -4,14 +4,18 @@ import (
 	"archive/tar"
 	"bytes"
 	"compress/gzip"
+	"fmt"
 	"io"
+	"math/rand"
 	"time"
 )
 
 type Archive struct {
```
rhcarvalho: @philbrookes could you document these types? Archive, ArchiveWriter. It's not very clear what is the responsibility of each of those to me.

If we have an "ArchiveWriter", shouldn't it be the one responsible for serializing writes?!

tgz.go
```diff
@@ -67,6 +96,10 @@ func (a *Archive) AddFileByContent(src []byte, dest string) error {
 }
 
 func (a *Archive) Close() {
+	close(a.WriteQueue)
```
rhcarvalho: Closing the channel at this point may lead to unintended panics.

Notice that calling `ArchiveWriter.Close` sends on the channel, and sending on a close channel panics.

Anyway, we shall fix this by not exposing the channel to multiple values (Archive, ArchiveWriter).

tgz_test.go
```diff
@@ -34,8 +35,148 @@ func TestWritingAFile(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if _, ok := files["logs/projects/phils-core/pods-fh-aaa-8-v7m10-fh-aaa.logs"]; !ok {
+		t.Fatal("Expected tgz to contain logs/projects/phils-core/pods-fh-aaa-8-v7m10-fh-aaa.logs but it didnt")
+	}
+}
+
+type task func() error
```
rhcarvalho: Why do we need this type? It is the same underlying type as `Task`.

tgz_test.go
```diff
@@ -34,8 +35,148 @@ func TestWritingAFile(t *testing.T) {
...
+	numSyncWrites := 25000
+
+	tasks := []task{}
+
+	for i := 1; i <= numSyncWrites; i++ {
+		tasks = append(tasks, getWriteTask(tgz.GetWriterToFile("path/to/file"+strconv.Itoa(i)+".file")))
+	}
+
+	var wg sync.WaitGroup
+	sem := make(chan struct{}, numSyncWrites)
```
rhcarvalho: The number of tasks and the size of the semaphore are the same, so the semaphore is not useful :)
This test function is creating 25,000 goroutines in any case.

tgz_test.go
```diff
@@ -34,8 +35,148 @@ func TestWritingAFile(t *testing.T) {
...
+	for _, task := range tasks {
+		task := task
+		wg.Add(1)
+		sem <- struct{}{}
+		go func() {
+			defer wg.Done()
+			task()
```
rhcarvalho: If we ignore the error, why do we even bother making that `task` type be a function that returns an error?
And the tasks always return a hard-coded `nil` (https://github.com/feedhenry/fh-system-dump-tool/pull/17/files#diff-b2fb05d71eb58d7fb88a0176b1b78bfcR127)

tgz_test.go
```diff
@@ -34,8 +35,148 @@ func TestWritingAFile(t *testing.T) {
...
+
+	tgz.Close()
+
+	files, err := decompressAndListFiles(&b)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if _, ok := files["path/to/file1.file"]; !ok {
+		t.Fatal("Expected tgz to contain path/to/file1.file but it didnt")
```
rhcarvalho: Here you don't need to call `t.Fatal`, rather `t.Error*` because it is fine and desirable to proceed and run the rest of the function body (to make other checks).

philbrookes:
Closed due to change in scope at: https://issues.jboss.org/browse/RHMAP-9412

PR #19, opened 2016-08-11
Update Travis config

rhcarvalho:
- Run checks with Go 1.6.x
- Run `go test` and `go build` in all packages
- Promote automatically gofmt'ing code instead of applying manual fixes based on diff output

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
- - tip
+ - 1.6.3
 script:
   - gofmt -d -s $(find . -name '*.go')
   - test -z "$(gofmt -d -s $(find . -name '*.go'))"
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,5 @@ language: go
 go:
  - 1.6.3
 script:
-  - gofmt -d -s $(find . -name '*.go')
-  - test -z "$(gofmt -d -s $(find . -name '*.go'))"
+  - test -z "$(gofmt -s -l $(find . -name '*.go'))" || { echo "Code needs to be gofmt'ed"; false; }
   - go test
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,9 +1,6 @@
-
 language: go
-go: 
- - tip
-
+go:
+ - 1.6.3
```
philbrookes: Didn't we agree to test both `tip` and `1.6.3`?

rhcarvalho:
Example of passing build:
https://travis-ci.org/feedhenry/fh-system-dump-tool/builds/151450542

Example of build showing how tip is slow (requires building Go itself):
https://travis-ci.org/feedhenry/fh-system-dump-tool/builds/151449259

Example of build showing how release-branch.go1.6 is slow (requires building Go itself):
https://travis-ci.org/feedhenry/fh-system-dump-tool/builds/151447772

Example of failing build (after I intentionally added a file violating gofmt, and a failing test):
https://travis-ci.org/feedhenry/fh-system-dump-tool/builds/151451508

.travis.yml
rhcarvalho: We did, but see:
- https://github.com/feedhenry/fh-system-dump-tool/pull/19#issuecomment-239110882
- https://travis-ci.org/feedhenry/fh-system-dump-tool/builds/151449259

Making Travis build Go tip to run our tests wastes our time.

We use 1.6.3 to build releases, that should be the version we test against IMHO.
I initially left both tip and 1.6.3 in there, but I think I prefer the fastest feedback of just 1.6.3.

Do you still see strong value in running all PRs against Go tip?

philbrookes: I see, I did not realize go was built for executing against tip, that seems excessive for less value. So in this case I'm happy to just test against 1.6.3.

philbrookes:
:+1:

PR #20, opened 2016-08-11
RHMAP-8114 add nagiosData dump task

philbrookes:
# Motivation
In order to successfully diagnose customer issues using the data dumped from their system the Nagios status and history wil lbe invaluable in helping us diagnose issues. This PR intends to include the nagios status data and historical data in the dump file, for every project it is available in.

# Changes
Add 2 new tasks to the dump tool - per project:
- one to dump the current status
- one to dump the historical status

### DIFF
--- a/logs.go
+++ b/logs.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"os/exec"
 	"strconv"
+	"strings"
 )
 
 // LoggableResource describes an OpenShift resource that produces logs.
@@ -89,6 +90,25 @@ func getLoggableResources(getPodContainers func(string, string) ([]string, error
 			})
 	}
 	return loggableResources, nil
+}
+
+// getResourceNamesBySubstr returns a list of names for the provided resource type that contain
+// the provided string, in the provided project.
+func getResourceNamesBySubstr(project, resource, substr string) ([]string, error) {
+	resources, err := GetResourceNames(project, resource)
+	if err != nil {
+		return nil, err
+	}
+
+	filtered := resources[:0]
+
+	for _, resource := range resources {
+		if strings.Contains(resource, substr) {
+			filtered = append(filtered, resource)
+		}
+	}
+
+	return filtered, nil
 }
 
 // GetPodContainers returns a list of container names for the named pod in the
--- a/nagios.go
+++ b/nagios.go
@@ -1,0 +1,55 @@
+package main
+
+import (
+	"fmt"
+	"os/exec"
+)
+
+// GetNagiosStatusData is a task factory for tasks that fetch a JSON representation of
+// the provided project. The task uses outFor and errOutFor to get io.Writers
+// to write, respectively, the JSON output and any eventual error message.
+func GetNagiosStatusData(project, pod string, outFor, errOutFor projectResourceWriterCloserFactory) Task {
+	return retrieveNagiosData(func() *exec.Cmd {
+		return exec.Command("oc", "exec", pod, "--", "cat", "/var/log/nagios/status.dat")
+	}, project, pod, "status", outFor, errOutFor)
+}
+
+// GetNagiosHistoricalData is a task factory for tasks that fetch a JSON representation of
+// the provided project. The task uses outFor and errOutFor to get io.Writers
+// to write, respectively, the JSON output and any eventual error message.
+func GetNagiosHistoricalData(project, pod string, outFor, errOutFor projectResourceWriterCloserFactory) Task {
+	return retrieveNagiosData(func() *exec.Cmd {
+		return exec.Command("oc", "exec", pod, "--", "tar", "-c", "-C", "/var/log/nagios", "archives")
+	}, project, pod, "history", outFor, errOutFor)
+}
+
+// A getNagiosDataCmdFactory generates commands to get the raw nagios status data
+// in the provided project from the provided pod
+type getNagiosDataCmdFactory func() *exec.Cmd
+
+// retrieveNagiosData will dump a JSON representation of the current nagios status into
+// the io.Writer returned from outFor and log any errors it encounters doing so
+// into the io.Writer returned from errOutFor.
+func retrieveNagiosData(cmdFactory getNagiosDataCmdFactory, project, pod, nagiosResourceType string, outFor, errOutFor projectResourceWriterCloserFactory) Task {
+	return func() error {
+		stdout, stdoutCloser, err := outFor(project, pod+"-"+nagiosResourceType)
+		if err != nil {
+			return err
+		}
+		defer stdoutCloser.Close()
+
+		stderr, stderrCloser, err := errOutFor(project, pod+"-"+nagiosResourceType)
+		if err != nil {
+			return err
+		}
+		defer stderrCloser.Close()
+
+		cmd := cmdFactory()
+
+		if err := runCmdCaptureOutput(cmd, stdout, stderr); err != nil {
+			fmt.Fprint(stderr, err)
+			return err
+		}
+		return nil
+	}
+}
--- a/tasks.go
+++ b/tasks.go
@@ -42,6 +42,12 @@ func GetAllTasks(basepath string) ([]Task, error) {
 	}
 	tasks = append(tasks, logsTasks...)
 
+	nagiosDataTasks, err := GetNagiosTasks(projects, basepath)
+	if err != nil {
+		retErrors = append(retErrors, err)
+	}
+	tasks = append(tasks, nagiosDataTasks...)
+
 	// Add check tasks
 	for _, p := range projects {
 		outFor := outToFile(basepath, "json", "analysis")
@@ -52,6 +58,34 @@ func GetAllTasks(basepath string) ([]Task, error) {
 
 	if len(retErrors) > 0 {
 		return tasks, retErrors
+	}
+	return tasks, nil
+}
+
+// GetNagiosTasks will return an array of Tasks each of which will dump the nagios data for one project.
+func GetNagiosTasks(projects []string, basepath string) ([]Task, error) {
+	var tasks []Task
+	var errors errorList
+	for _, p := range projects {
+		pods, err := getResourceNamesBySubstr(p, "pod", "nagios")
+		if err != nil {
+			errors = append(errors, err)
+			continue
+		}
+		for _, pod := range pods {
+			outFor := outToFile(basepath, "dat", "nagios")
+			errOutFor := outToFile(basepath, "stderr", "nagios")
+			task := GetNagiosStatusData(p, pod, outFor, errOutFor)
+			tasks = append(tasks, task)
+
+			outFor = outToFile(basepath, "tar", "nagios")
+			errOutFor = outToFile(basepath, "stderr", "nagios")
+			task = GetNagiosHistoricalData(p, pod, outFor, errOutFor)
+			tasks = append(tasks, task)
+		}
+	}
+	if len(errors) > 0 {
+		return tasks, errors
 	}
 	return tasks, nil
 }


logs.go
```diff
@@ -91,6 +93,32 @@ func getLoggableResources(getPodContainers func(string, string) ([]string, error
...
+
+	if len(ret) == 0 {
+		return nil, errors.New("No pods in '" + project + "' matched pod substring '" + substr + "'")
+	}
+
+	return ret, nil
+}
+
+// GetProjectPods returns a list of pod names for the named project.
+func getProjectPodNames(project string) ([]string, error) {
```
rhcarvalho: This is the same as `GetResourceNames(project, "pod")`

rhcarvalho:
@philbrookes file names should be lowercase in Go, s/nagiosData.go/nagios.go/

nagiosData.go
```diff
@@ -0,0 +1,103 @@
...
+
+		stdErr, stdErrCloser, err := errOutFor(project, "nagios")
+		if err != nil {
+			return err
+		}
+		defer stdErrCloser.Close()
+
+		stdOutBuff := bytes.NewBuffer([]byte{})
+
+		nagiosPod, err := getProjectPodNamesBySubstr(project, "nagios")
```
rhcarvalho: Is it reasonable to do this and then take `nagiosPod[0]`?

nagiosData.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+		defer stdOutCloser.Close()
+
+		stdErr, stdErrCloser, err := errOutFor(project, "nagios")
+		if err != nil {
+			return err
+		}
+		defer stdErrCloser.Close()
+
+		stdOutBuff := bytes.NewBuffer([]byte{})
```
rhcarvalho:
```go
var b bytes.Buffer
```

nagiosData.go
```diff
@@ -0,0 +1,103 @@
...
+			return err
+		}
+		cmd := cmdFactory(project, nagiosPod[0])
+
+		if err := runCmdCaptureOutput(cmd, stdOutBuff, stdErr); err != nil {
+			stdErr.Write([]byte(err.Error()))
+			return err
+		}
+
+		stdOut.Write(parseNagiosStatus(string(stdOutBuff.Bytes())))
```
rhcarvalho: This is where a MultiWriter is handy... see https://github.com/feedhenry/fh-system-dump-tool/blob/ab1274f54d88f2216de69ea547d65a1fc2ce7c92/main.go#L49

nagiosData.go
```diff
@@ -0,0 +1,103 @@
...
+		if err != nil {
+			return err
+		}
+		defer stdErrCloser.Close()
+
+		stdOutBuff := bytes.NewBuffer([]byte{})
+
+		nagiosPod, err := getProjectPodNamesBySubstr(project, "nagios")
+		if err != nil {
+			stdErr.Write([]byte(err.Error()))
```
rhcarvalho:
```go
fmt.Fprint(stderr, err)
```

nagiosData.go
philbrookes: @rhcarvalho I'm not understanding how a multiwriter is useful here. I am passing the value that was written into the `stdoutBuff` value as a string to the `parseNagiosStatus` function, and writing the result of that parsing into the `stdOut` writer.

nagios.go
```diff
@@ -0,0 +1,113 @@
...
+
+		var stdoutBuff bytes.Buffer
+
+		nagiosPod, err := getProjectPodNamesBySubstr(project, "nagios")
+		if err != nil {
+			fmt.Fprint(stderr, err)
+			return err
+		}
+
+		if len(nagiosPod) == 0 {
```
philbrookes: Moved this check out of the `getProjectPodNamesBySubstr` I think it makes more sense up here.

nagios.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+
+		cmd := cmdFactory(project, nagiosPod[0])
+
+		if err := runCmdCaptureOutput(cmd, &stdoutBuff, stderr); err != nil {
+			stderr.Write([]byte(err.Error()))
+			return err
+		}
+
+		jsonStatus := parseNagiosStatus(string(stdoutBuff.Bytes()))
```
philbrookes: Split this into two lines, for added clarity.

nagios.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+
+		cmd := cmdFactory(project, nagiosPod[0])
+
+		if err := runCmdCaptureOutput(cmd, &stdoutBuff, stderr); err != nil {
+			fmt.Fprint(stderr, err)
+			return err
+		}
+
+		jsonStatus := parseNagiosStatus(string(stdoutBuff.Bytes()))
```
rhcarvalho: I see what you meant, the catchy thing is `parseNagiosStatus` parsing an arbitrary string.
Isn't there a "Nagios API"? /cc @jessesarn

nagios.go
```diff
@@ -0,0 +1,113 @@
...
+			continue
+		}
+
+		if strings.TrimSpace(line) == "}" {
+			conf[currentId] = append(conf[currentId], curConf)
+			continue
+		}
+	}
+
+	retVal, _ := json.MarshalIndent(conf, "", " ")
```
rhcarvalho: Why ignore the error?

nagios.go
```diff
@@ -0,0 +1,113 @@
...
+	}, project, outFor, errOutFor)
+}
+
+// A getNagiosDataCmdFactory generates commands to get the raw nagios status data
+// in the provided project from the provided pod
+type getNagiosDataCmdFactory func(project, pod string) *exec.Cmd
+
+// nagiosData will dump a JSON representation of the current nagios status into
+// the io.Writer returned from outFor and log any errors it encounters doing so
+// into the io.Writer returned from errOutFor.
```
rhcarvalho: Why do we even bother converting this to JSON, instead of storing in whatever format Nagios can serve us?

nagios.go
philbrookes: There is a nagios api, but it's a third party tool that runs it's own daemon, and essentially does exactly what I'm doing here (i.e. parsing the status.dat file into json). I discussed this with @david-martin and we thought this was less intrusive than creating a new container for the nagios-api.

nagios.go
philbrookes: I'm not sure why I chose to ignore this, I assume it's an oversight during development so I'll add some code to handle it.

nagios.go
philbrookes: The format that nagios uses is a custom nagios format, I don't have a particular problem with dumping in that format but in that case, I think then we'd end up doing this parsing in any checks that read this file instead. So it is probably easier to just parse it to JSON pre-write instead?

rhcarvalho: We could accidentally introduce errors here. I'd opt for keeping the "run a command and save the output", that will give us the Nagios logs in its original format.

Yes we might need to parse that later as part of the analysis, but I'd suggest we do it then, not in the dump part.
Save the code, it will be useful soon :)


BTW, is this .dat file an ever growing file with everything since the first run?

nagios.go
rhcarvalho: Yep, let's keep it simple.

nagios.go
jessesarn: Yeah, I agree, we should avoid processing the raw data during the dump phase.  A mistake in the translation logic could result in us not getting the data we may need in the event that our analysis isn't able to determine a problem.  I think the rawer, the data, the better :)

philbrookes: Ok, I'm convinced :) No the file doesn't grow, it is rewritten when nagios checks are updated.

rhcarvalho: > No the file doesn't grow, it is rewritten when nagios checks are updated.

And do you know where we can get "Nagios historical data" from? @philbrookes @jessesarn

jessesarn: We have a tool written for Ops that queries for the historical info from:
```
for day in range(dayStart, dayEnd)
  url = nagUrl + '/cgi-bin/nagios3/history.cgi?host=all&type=0&statetype=0&archive=' + str(day)
```

philbrookes: @jessesarn @rhcarvalho this is a bit awkward, I have been able to programmatically retrieve the route for the nagios end-point, but then realised that I'll also need the http auth details to get into the URL which feels like something of a blocker.

rhcarvalho: We can take files from disk, not use HTTP calls, that's what I understood before.

philbrookes: Looking a little further into this, it seems that those archive pages are generated based on the files in that archive dir, so I will just dump those raw into the dump.

philbrookes: @rhcarvalho What would you see as the best way to dump all files in a directory? I'm currently thinking of doing an ls of the dir, then a cat on each file, but that seems a bit messy?

rhcarvalho: Use `oc exec`, `tar` the dir to stdout and pipe to tar it out into the output dir.

philbrookes: @rhcarvalho for me to do it this way - unless I've missed something - I cannot send the output down the stdout writer, I will instead need the task to be aware of the path to the directory it should be writing to?

philbrookes: @rhcarvalho Maybe the best solution here is to have the task write to a .tar file and dump the output of the tar file straight into that?

nagiosData.go
vchepeli: @philbrookes io.MultiWriter(parseNagiosStatus(string(stdOutBuff.Bytes())), stdOut) or even much more readable. Here nobody knows what you read from and what you write to StdOut

nagios.go
vchepeli: Whats nagiosPod here? Array or string?

logs.go
```diff
@@ -91,6 +92,23 @@ func getLoggableResources(getPodContainers func(string, string) ([]string, error
 	return loggableResources, nil
 }
 
+// GetProjectPodsBySubstr returns a list of pod names that contain the provided string, in the named project.
+func getProjectPodNamesBySubstr(project, substr string) ([]string, error) {
+	ret := []string{}
+	pods, err := GetResourceNames(project, "pod")
```
vchepeli: "pod" should be constant

logs.go
```diff
@@ -91,6 +92,23 @@ func getLoggableResources(getPodContainers func(string, string) ([]string, error
...
+		return nil, err
+	}
+
+	for _, pod := range pods {
+		if strings.Contains(pod, substr) {
+			ret = append(ret, pod)
+		}
+	}
+
+	return ret, nil
```
vchepeli: why we return **nil** here?

nagios.go
philbrookes: @vchepeli an array of strings.

logs.go
philbrookes: @vchepeli At this point the function has executed successfully, so this returns nil in the error part of the return.

nagios.go
```diff
@@ -0,0 +1,113 @@
...
+	"os/exec"
+	"regexp"
+	"strings"
+)
+
+// NagiosData is a task factory for tasks that fetch a JSON representation of
+// the provided project. The task uses outFor and errOutFor to get io.Writers
+// to write, respectively, the JSON output and any eventual error message.
+func NagiosData(project string, outFor, errOutFor projectResourceWriterCloserFactory) Task {
+	return nagiosData(func(project, pod string) *exec.Cmd {
```
vchepeli: @philbrookes you pass anonymous function here? Why just not create function runCmd() with two args <command, path/to/log/file>

vchepeli: @philbrookes Is this function a clousure? projectResourceWriterCloserFactory is not used anywhere?

philbrookes: @vchepeli the `outFor` and `errOutFor` are passed as arguments into `nagiosData` here:
https://github.com/philbrookes/fh-system-dump-tool/blob/RHMAP-8114/nagios.go#L19

The anonymous function seems sensible to me, as it is such a specific use case, I don't see what we'd gain by defining the function outside and calling it here?

nagios.go
vchepeli: @philbrookes then name should be **nagiosPods**

logs.go
vchepeli: @philbrookes can we return just empty error Error.NIL and not nil?

nagios.go
vchepeli: Then make it named if its sensible for you :)

nagios.go
philbrookes: @vchepeli good point thanks, I'll update that.

nagios.go
vchepeli: Here I agree with @rhcarvalho. We should keep raw data and not convert it to JSON. Converting to JSON or another format should be next level

nagios.go
vchepeli: @philbrookes and len(nagiosPod) == 0 is ugly. We need to change it to if(nagiosPods.isEmpty)

nagios.go
```diff
@@ -0,0 +1,113 @@
...
+
+		if err := runCmdCaptureOutput(cmd, &stdoutBuff, stderr); err != nil {
+			fmt.Fprint(stderr, err)
+			return err
+		}
+
+		jsonStatus := parseNagiosStatus(string(stdoutBuff.Bytes()))
+		stdout.Write(jsonStatus)
+
+		return nil
```
vchepeli: @philbrookes what is nil here? That is error prone too? :))

nagios.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+}
+
+// parseNagiosStatus takes the raw nagios status string and converts
+// it into a string of JSON
+func parseNagiosStatus(raw string) []byte {
```
vchepeli: This function should be rewritten! It has 50+ lines. Divide into smaller 5-6 with max 10 lines of code

nagios.go
vchepeli: @philbrookes all errors should be handled and formatted! But dont forget to check your **nil**s too

tasks.go
```diff
@@ -42,6 +42,12 @@ func GetAllTasks(basepath string) ([]Task, error) {
...
 	tasks = append(tasks, logsTasks...)
 
+	nagiosDataTasks, err := GetNagiosDataTask(projects, basepath)
+	if err != nil {
+		retErrors = append(retErrors, err)
+	}
+	tasks = append(tasks, nagiosDataTasks...)
+
 	// Add check tasks
 	for _, p := range projects {
```
vchepeli: What you ignore here?

tasks.go
```diff
@@ -56,6 +62,18 @@ func GetAllTasks(basepath string) ([]Task, error) {
 	return tasks, nil
 }
 
+// GetNagiosDataTask will return an array of Tasks each of which will dump the nagios data for one project.
+func GetNagiosDataTask(projects []string, basepath string) ([]Task, error) {
+	var tasks []Task
+	for _, p := range projects {
+		outFor := outToFile(basepath, "json", "logs")
+		errOutFor := outToFile(basepath, "stderr", "logs")
```
vchepeli: All this "<string-constant>" should be constants!

logs.go
vchepeli: @philbrookes and we should not care about what error it returns it should be passed to caller function! And caller function should check for possible errors

vchepeli:
VD, comments addressed, motivation is not present for this PR, take inspiration from [here](https://github.com/feedhenry/gitlab-shell/pull/17)

nagios.go
philbrookes: @vchepeli as far as I know, go-lang does not have a method like `isEmpty` for arrays / slices and `len(slice)` is the standard way to check for this case. @rhcarvalho @maleck13 Could I get your opinion on this?

logs.go
rhcarvalho: @vchepeli I suggest https://tour.golang.org and later https://golang.org/doc/effective_go.html to get familiar with the language.

nagios.go
philbrookes: @vchepeli I'm not sure what you mean here, if you are suggesting changing this to:
```
return nagiosData(func SomeFunctionNameHere(project, pod string) *exec.Cmd {
```
Then that is not valid syntax for go.

I can change it to:
```
        cmdFactory := func(project, pod string) *exec.Cmd {
		return exec.Command("oc", "exec", pod, "--", "cat", "/var/log/nagios/status.dat")
	}
	return nagiosData(cmdFactory, project, "Status", outFor, errOutFor)
```

If you find that syntax preferable, I will change it to that.

logs.go
philbrookes: @vchepeli There is no `Error.NIL` in go, returning `nil` for errors when no error has occurred is a pretty common usage, see here for more info: https://golang.org/doc/effective_go.html#multiple-returns

nagios.go
philbrookes: @vchepeli this is the same as above, this return of `nil` is to signify that no error has occurred.

nagios.go
rhcarvalho: > Whats nagiosPod here? Array or string?

To be precise it is neither of the two. It is a `[]string`, a.k.a "slice of string".

> @philbrookes and len(nagiosPod) == 0 is ugly. We need to change it to if(nagiosPods.isEmpty) cause in this context I dont know if that is string or array

There's no such a thing. @philbrookes's got it right, the Go way.

---

About the name, it is just a sign of a bigger dilemma. We "grep" something in search for the Nagios pod and we hope there is only one -- not zero, and not more.
The first part we're checking for, the second part we are just assuming that the first one is the one we care about (when we write below `nagiosPod[0]`).

nagios.go
rhcarvalho: Please don't change anything here. That's simply the type of the argument to the function we are calling.

philbrookes:
@vchepeli I am not sure what VD stands for, could you clarify please?

logs.go
philbrookes: Although I understand the case when a constant can be useful, I do not think this should be a blanket rule that is applied to any literal string. 

Particularly in this cases like this, I do not see any advantage to having a `constant POD = "pod"` and changing this to `GetResourceNames(project, POD)`

philbrookes:
I have removed the parseNagiosStatus function as it is not currently in use, we will likely need it in future, so I've dumped it here: https://gist.github.com/philbrookes/2bd0d13003c7767d658561846f267bb5

david-martin:
@vchepeli Thanks for chiming in here.
I agree that having the motivation and changes is useful in a PR (similar to the link you posted).
At the very least there should be a jira issue to fall back on.
I think we're good here now.

I'll defer to @rhcarvalho and @philbrookes on any other comments you left as they are more familiar with golang and this tool than I am.

nagios.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		defer stderrCloser.Close()
+
+		nagiosPods, err := getProjectPodNamesBySubstr(project, nagiosName)
+		if err != nil {
+			fmt.Fprint(stderr, err)
+			return err
+		}
+
+		if len(nagiosPods) == 0 {
```
rhcarvalho: Maybe this should not be an error, at least not in the current form of the code, because we are iterating over all projects and looking for a "nagios" pod in there?

logs.go
```diff
@@ -91,6 +92,23 @@ func getLoggableResources(getPodContainers func(string, string) ([]string, error
 	return loggableResources, nil
 }
 
+// GetProjectPodsBySubstr returns a list of pod names that contain the provided string, in the named project.
+func getProjectPodNamesBySubstr(project, substr string) ([]string, error) {
```
rhcarvalho: Wouldn't it be better if instead of this function we would try to find a `*nagios*` pod in all `projects` and print an error (in main.go) in case none was found?

tasks.go
```diff
@@ -56,6 +62,24 @@ func GetAllTasks(basepath string) ([]Task, error) {
 	return tasks, nil
 }
 
+// GetNagiosDataTask will return an array of Tasks each of which will dump the nagios data for one project.
+func GetNagiosTasks(projects []string, basepath string) ([]Task, error) {
+	var tasks []Task
+	for _, p := range projects {
```
rhcarvalho: Here is the place where instead of creating as many tasks as projects, I'd find the 1 Nagios we are looking for, and then create 1 task.

logs.go
vchepeli: @rhcarvalho here unfortunately I don't care about any language. You can make mess in any language. I take care about QoC and if it's nil/null, we should know why we return it

nagios.go
vchepeli: @philbrookes why we just not do only one return err; and not doing it to different cases and if err happens we pass it to caller function?

logs.go
vchepeli: @philbrookes why we just not do only onereturn ret, err and not doing it to different cases and if err happens we pass it to caller function?

vchepeli:
@philbrookes VD means Verification Done :)) or Review Done

nagios.go
```diff
@@ -0,0 +1,74 @@
...
+		return exec.Command("oc", "exec", pod, "--", "cat", "/var/log/nagios/status.dat")
+	}, project, "Status", outFor, errOutFor)
+}
+
+// NagiosStatusData is a task factory for tasks that fetch a JSON representation of
+// the provided project. The task uses outFor and errOutFor to get io.Writers
+// to write, respectively, the JSON output and any eventual error message.
+func GetNagiosHistoricalData(project string, outFor, errOutFor projectResourceWriterCloserFactory) Task {
+	return retrieveNagiosData(func(project, pod string) *exec.Cmd {
+		return exec.Command("oc", "exec", pod, "--", "tar", "-c", "-C", "/var/log/nagios", "archives", "--to-stdout")
```
rhcarvalho: Do we want to keep the full path in the archive?

If we would untar the output, would we want it to be in somewhere like
a. `rhmap-dumps/TIMESTAMP/nagios/var/log/nagios/...`; or
b. `rhmap-dumps/TIMESTAMP/nagios/...`

a or b?

nagios.go
```diff
@@ -0,0 +1,74 @@
...
+	"os/exec"
+)
+
+// NagiosStatusData is a task factory for tasks that fetch a JSON representation of
+// the provided project. The task uses outFor and errOutFor to get io.Writers
+// to write, respectively, the JSON output and any eventual error message.
+func GetNagiosStatusData(project string, outFor, errOutFor projectResourceWriterCloserFactory) Task {
+	return retrieveNagiosData(func(project, pod string) *exec.Cmd {
+		return exec.Command("oc", "exec", pod, "--", "cat", "/var/log/nagios/status.dat")
+	}, project, "Status", outFor, errOutFor)
```
rhcarvalho: Why is `Status` capitalized?

logs.go
philbrookes: @vchepeli The philosophy here is to exit early, rather than saving up errors until the end of the function. If an error is discovered then it should be reported and the function should exit because if an error is found it would be unwise to keep going in a known to be unstable situation.

To keep these advantages and only have one return statement would mean nesting if statements around every possible error scenario, which is much less readable.

nagios.go
philbrookes: I have gone with B here, which is what the `-C` is doing, if we'd prefer A then I can remove that argument, any thoughts on this @david-martin @pmccarthy @jessesarn?

philbrookes:
@rhcarvalho Assuming the path inside the dump directory is OK. Does everything else looks OK to merge here?

logs.go
```diff
@@ -91,6 +92,23 @@ func getLoggableResources(getPodContainers func(string, string) ([]string, error
...
+	ret := []string{}
```
rhcarvalho: https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

And if we are to keep this function, we could organize it better to address the readability concerns:

```go
pods, err := GetResourceNames(project, "pod")
if err != nil {
	return nil, err
}
var filtered []string
for _, pod := range pods {
// ...
}
return filtered, nil
```

We only need `filtered` past some point into the function body. And perhaps `filtered` helps understand the intent better than `ret`?

And since we are talking about filtering, sharing a trick:
https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating

Anyway, I think this function would be easier to use if it didn't know about pods... you just need to filter an arbitrary slice of strings.

nagios.go
```diff
@@ -0,0 +1,55 @@
...
+		return exec.Command("oc", "exec", pod, "--", "cat", "/var/log/nagios/status.dat")
+	}, project, pod, "status", outFor, errOutFor)
+}
+
+// NagiosStatusData is a task factory for tasks that fetch a JSON representation of
+// the provided project. The task uses outFor and errOutFor to get io.Writers
+// to write, respectively, the JSON output and any eventual error message.
+func GetNagiosHistoricalData(project, pod string, outFor, errOutFor projectResourceWriterCloserFactory) Task {
+	return retrieveNagiosData(func() *exec.Cmd {
+		return exec.Command("oc", "exec", pod, "--", "tar", "-c", "-C", "/var/log/nagios", "archives", "--to-stdout")
```
rhcarvalho: According to `man tar`:

```
    -O, --to-stdout
              Extract files to standard output.
```

It doesn't seen to have any relation to creating archives?

In a quick local test `--to-stdout` seems to have no influence on `tar c foo > foo.tar`.

nagios.go
```diff
@@ -0,0 +1,55 @@
...
+	}, project, pod, "history", outFor, errOutFor)
+}
+
+// A getNagiosStatusDataCmdFactory generates commands to get the raw nagios status data
```
rhcarvalho: Name mismatch

nagios.go
```diff
@@ -0,0 +1,55 @@
...
+// in the provided project from the provided pod
+type getNagiosDataCmdFactory func() *exec.Cmd
+
+// nagiosData will dump a JSON representation of the current nagios status into
```
rhcarvalho: Name mismatch again :)

nagios.go
```diff
@@ -0,0 +1,55 @@
+package main
+
+import (
+	"fmt"
+	"os/exec"
+)
+
+// NagiosStatusData is a task factory for tasks that fetch a JSON representation of
```
rhcarvalho: wrong name

nagios.go
```diff
@@ -0,0 +1,55 @@
...
+// the provided project. The task uses outFor and errOutFor to get io.Writers
+// to write, respectively, the JSON output and any eventual error message.
+func GetNagiosStatusData(project, pod string, outFor, errOutFor projectResourceWriterCloserFactory) Task {
+	return retrieveNagiosData(func() *exec.Cmd {
+		return exec.Command("oc", "exec", pod, "--", "cat", "/var/log/nagios/status.dat")
+	}, project, pod, "status", outFor, errOutFor)
+}
+
+// NagiosStatusData is a task factory for tasks that fetch a JSON representation of
```
rhcarvalho: wrong name

logs.go
```diff
@@ -91,6 +92,23 @@ func getLoggableResources(getPodContainers func(string, string) ([]string, error
 	return loggableResources, nil
 }
 
+// GetProjectPodsBySubstr returns a list of pod names that contain the provided string, in the named project.
```
rhcarvalho: name doesn't match func name

tasks.go
```diff
@@ -56,6 +62,32 @@ func GetAllTasks(basepath string) ([]Task, error) {
...
+		pods, err := getResourceNamesBySubstr(p, "pod", "nagios")
+		if err != nil {
+			continue
```
rhcarvalho: Why ignore the error completely?

tasks.go
```diff
@@ -56,6 +62,32 @@ func GetAllTasks(basepath string) ([]Task, error) {
...
+		}
+		if len(pods) == 0 {
```
rhcarvalho: This is useless, iterating over the empty slice has the same effect.

tasks.go
```diff
@@ -56,6 +62,32 @@ func GetAllTasks(basepath string) ([]Task, error) {
...
+			continue
+		}
+		for _, pod := range pods {
+			outFor := outToFile(basepath, "dat", "logs")
```
rhcarvalho: Why are we saving to "logs"?

That will mix pod logs and nagios data?!

philbrookes: I felt like the status and historical data of Nagios was like a log of what nagios has done. What would you recommend here, a new folder like "Monitoring"?

tasks.go
```diff
@@ -56,6 +62,33 @@ func GetAllTasks(basepath string) ([]Task, error) {
 	return tasks, nil
 }
 
+// GetNagiosDataTask will return an array of Tasks each of which will dump the nagios data for one project.
```
rhcarvalho: name mismatch

tasks.go
```diff
@@ -56,6 +62,33 @@ func GetAllTasks(basepath string) ([]Task, error) {
...
+func GetNagiosTasks(projects []string, basepath string) ([]Task, error) {
+	var tasks []Task
+	var errors errorList
+	for _, p := range projects {
+		pods, err := getResourceNamesBySubstr(p, "pod", "nagios")
+		if err != nil {
+			errors = append(errors, err)
```
rhcarvalho: Does it make sense to loop over pods when there was an error from `getResourceNamesBySubstr`?

philbrookes: no it doesn't. I've added a continue after appending this error now.

tasks.go
philbrookes: fixed now.

rhcarvalho:
LGTM

tasks.go
```diff
@@ -56,6 +62,34 @@ func GetAllTasks(basepath string) ([]Task, error) {
...
+			continue
+		}
+		for _, pod := range pods {
+			outFor := outToFile(basepath, "dat", "nagios")
+			errOutFor := outToFile(basepath, "stderr", "nagios")
+			task := GetNagiosStatusData(p, pod, outFor, errOutFor)
+			tasks = append(tasks, task)
+
+			outFor = outToFile(basepath, "tar", "nagios")
+			errOutFor = outToFile(basepath, "stderr", "nagios")
```
rhcarvalho: @philbrookes I suspect this overwrites the stderr for the previous command?

rhcarvalho: Hmm, I think it doesn't, the "scope" is the same, but the file names will be "status.stderr" and "history.stderr".

philbrookes: yes, this is correct, the first set of tasks create:
```
nagios/<pod>-status.dat
nagios/<pod>-status.stderr
```
The second set of tasks create:
```
nagios/<pod>-history.dat
nagios/<pod>-history.stderr
```
But that is decided in the `getNagios(Status|History)Data` function in the nagios.go file.

nagios.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+	"os/exec"
+)
+
+// GetNagiosStatusData is a task factory for tasks that fetch a JSON representation of
+// the provided project. The task uses outFor and errOutFor to get io.Writers
+// to write, respectively, the JSON output and any eventual error message.
+func GetNagiosStatusData(project, pod string, outFor, errOutFor projectResourceWriterCloserFactory) Task {
+	return retrieveNagiosData(func() *exec.Cmd {
+		return exec.Command("oc", "exec", pod, "--", "cat", "/var/log/nagios/status.dat")
```
rhcarvalho: @philbrookes shouldn't we be specifying also the container name here?

Came across this when refactoring, and wondering if it's not hiding the same situation with hit with logs.

nagios.go
```diff
@@ -0,0 +1,55 @@
...
+	return retrieveNagiosData(func() *exec.Cmd {
+		return exec.Command("oc", "exec", pod, "--", "tar", "-c", "-C", "/var/log/nagios", "archives")
+	}, project, pod, "history", outFor, errOutFor)
+}
+
+// A getNagiosDataCmdFactory generates commands to get the raw nagios status data
+// in the provided project from the provided pod
+type getNagiosDataCmdFactory func() *exec.Cmd
+
+// retrieveNagiosData will dump a JSON representation of the current nagios status into
```
rhcarvalho: This documentation is outdated, right? We are not producing JSON here, I think.

PR #21, opened 2016-08-12
Fetch log only for pods

rhcarvalho:
Deployment configs and other types are just alias to pods.

Through https://github.com/openshift/origin/issues/10186 we realized that `oc logs dc/...` will print application logs and not deployment logs after the deployment is completed and the deployer pod is deleted.

Thanks @jhellar!

### DIFF
--- a/logs.go
+++ b/logs.go
@@ -6,15 +6,19 @@ import (
 	"strconv"
 )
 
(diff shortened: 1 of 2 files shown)


rhcarvalho:
Simple enough, merging.

PR #22, opened 2016-08-15
Minor improvements

rhcarvalho:
Fixed a few nits:

- Usage of `rm` binary instead of the Go equivalent;
- Printing log message with `fmt.Print` instead log `log.Print`;
- Tool usage message consistency;
- Remove custom code when flag `-p 1` is passed (bug prone);

Improvement:

- Save all logs printed during tool execution as part of the dump.

### DIFF
--- a/main.go
+++ b/main.go
@@ -187,16 +187,6 @@ func main() {
 
 	log.Println("Running tasks...")
 
-	// Avoid the creating goroutines and other controls if we're executing
-	// tasks sequentially.
-	if *maxParallelTasks == 1 {
-		for _, task := range tasks {
-			task()
-			fmt.Fprint(os.Stderr, ".")
-		}
-		return
-	}
-
 	// Run at most N tasks in parallel, and wait for all of them to
 	// complete.
 	var wg sync.WaitGroup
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -175,24 +190,18 @@ func main() {
...
+		os.RemoveAll(basePath)
 
 		log.Printf("Dumped system information to: %s", basePath+".tar.gz")
 	}()
 
 	log.Println("Running tasks...")
 
-	// Avoid the creating goroutines and other controls if we're executing
-	// tasks sequentially.
-	if *maxParallelTasks == 1 {
```
philbrookes: Am I right in thinking this was removed as if maxParallelTasks was set to 1, it will do it sequentially in the wait group anyway?

philbrookes:
:+1: LGTM

main.go
rhcarvalho: Yes, the special case was there as a "safe way to run the tool without the concurrency stuff", but instead of doing good it was just accumulating bugs.
The number of parallel tasks defaults to the number of CPUs, so I've mostly been testing the tool with >1 values.

PR #23, opened 2016-08-15
Refactor task execution

rhcarvalho:
- Only create as many goroutines as needed, distribute tasks using a
  channel.
- Prepare tasks concurrently with task execution.

It also fixes an oversight: we were not handling task errors at all.

### DIFF
--- a/main.go
+++ b/main.go
@@ -13,7 +13,6 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
-	"sync"
 	"time"
 )
 
@@ -33,9 +32,9 @@ const (
 )
 
 var (
-	maxParallelTasks = flag.Int("p", runtime.NumCPU(), "max number of tasks to run in parallel")
-	maxLogLines      = flag.Int("max-log-lines", defaultMaxLogLines, "max number of log lines fetched with oc logs")
-	printVersion     = flag.Bool("version", false, "print version and exit")
+	concurrentTasks = flag.Int("p", runtime.NumCPU(), "number of tasks to run concurrently")
+	maxLogLines     = flag.Int("max-log-lines", defaultMaxLogLines, "max number of log lines fetched with oc logs")
+	printVersion    = flag.Bool("version", false, "print version and exit")
 )
 
 func runCmdCaptureOutput(cmd *exec.Cmd, out, errOut io.Writer) error {
@@ -140,7 +139,7 @@ func main() {
 		os.Exit(0)
 	}
 
-	if !(*maxParallelTasks > 0) {
+	if !(*concurrentTasks > 0) {
 		printError(fmt.Errorf("argument to -p flag must be greater than 0"))
 		os.Exit(1)
 	}
@@ -161,19 +160,6 @@ func main() {
 	defer logfile.Close()
 	log.SetOutput(io.MultiWriter(os.Stderr, logfile))
 	fileOnlyLogger := log.New(logfile, "", log.LstdFlags)
-
-	log.Print("Starting RHMAP System Dump Tool...")
-	log.Print("Preparing tasks...")
-
-	tasks, err := GetAllTasks(basePath)
-	if err != nil {
-		printError(err)
-		defer os.Exit(1)
-	}
-	if len(tasks) == 0 {
-		log.Print("No tasks found to execute.")
-		return
-	}
 
 	// defer creating a tar.gz file from the dumped output files
 	defer func() {
@@ -200,23 +186,8 @@ func main() {
 		log.Printf("Dumped system information to: %s", basePath+".tar.gz")
 	}()
 
-	log.Println("Running tasks...")
+	log.Print("Starting RHMAP System Dump Tool...")
+	log.Print("Running tasks...")
 
-	// Run at most N tasks in parallel, and wait for all of them to
-	// complete.
-	var wg sync.WaitGroup
-	sem := make(chan struct{}, *maxParallelTasks)
-	for _, task := range tasks {
-		task := task
-		wg.Add(1)
-		sem <- struct{}{}
-		go func() {
-			defer wg.Done()
-			task()
-			fmt.Fprint(os.Stderr, ".")
-			<-sem
-		}()
-	}
-	wg.Wait()
-	fmt.Fprintln(os.Stderr)
+	RunAllTasks(basePath, *concurrentTasks)
 }
--- a/tasks.go
+++ b/tasks.go
@@ -1,125 +1,162 @@
 package main
 
-import "errors"
+import (
+	"errors"
+	"fmt"
+	"log"
+	"os"
+	"sync"
+)
 
 // A Task performs some part of the RHMAP System Dump Tool.
 type Task func() error
 
-// GetAllTasks returns a list of all tasks performed by the dump tool. It may
-// return tasks even in the presence of an error.
-// FIXME: GetAllTasks should not need to know about basepath.
-func GetAllTasks(basepath string) ([]Task, error) {
-	var (
-		tasks     []Task
-		retErrors errorList
-	)
+// RunAllTasks runs all tasks known to the dump tool using concurrent workers.
+// Dump output goes to path.
+func RunAllTasks(path string, workers int) {
+	tasks := GetAllTasks(path)
+	results := make(chan error)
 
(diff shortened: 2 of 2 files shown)


main.go
```diff
@@ -200,23 +193,33 @@ func main() {
...
-	sem := make(chan struct{}, *maxParallelTasks)
-	for _, task := range tasks {
-		task := task
-		wg.Add(1)
-		sem <- struct{}{}
+	wg.Add(*parallelTasks)
+	for i := 0; i < *parallelTasks; i++ {
 		go func() {
-			defer wg.Done()
-			task()
```
rhcarvalho: @philbrookes notice here that we used to completely ignore task errors!
You can see below that now we log errors to the screen and to the "dump.log" that is part of the archive we produce.

tasks.go
```diff
@@ -1,125 +1,121 @@
...
+		wg.Wait()
+	}()
+	return tasks
 }
 
 // GetNagiosTasks will return an array of Tasks each of which will dump the nagios data for one project.
-func GetNagiosTasks(projects []string, basepath string) ([]Task, error) {
-	var tasks []Task
-	var errors errorList
+func GetNagiosTasks(projects []string, basepath string, tasks chan<- Task) {
```
rhcarvalho: The changes to the various functions here should be similar.

In essence, instead of returning `[]Task` and `error`, we take a `chan<- Task` as input.
We send tasks through the channel and they can be immediately run, as opposed to accumulating tasks in a list.

Given that running a task returns an error, and we often proceed even in the presence of errors, instead of taking a channel of errors as input, we simply send errors down the task channel as tasks that "always fail".

Having a single channel in this case facilitates synchronization.

tasks.go
```diff
@@ -1,125 +1,121 @@
...
 
-	// Add tasks to fetch logs.
-	logsTasks, err := GetFetchLogsTasks(projects, resourcesWithLogs, basepath)
-	if err != nil {
-		retErrors = append(retErrors, err)
-	}
-	tasks = append(tasks, logsTasks...)
+// GetAllTasks returns a list of all tasks known to the dump tool.
+// FIXME: GetAllTasks should not need to know about basepath.
+func GetAllTasks(basepath string) <-chan Task {
```
rhcarvalho: The big change to `GetAllTasks` can be summarized as:

Instead of sequentially accumulating lists of tasks from the various "task factories", we create a task channel and arrange for all factories to send tasks to it concurrently.

When all factories are done (`wg.Wait()` returns), we close the task channel. This communicates to the main function in main.go that no more tasks will be sent on the channel.

rhcarvalho: Note that the channel is only closed when all goroutines started here have terminated. And they terminate only after the tasks they produced have been received by one of the task runner goroutines from main (because the channel is unbuffered).

main.go
```diff
@@ -200,23 +193,33 @@ func main() {
 		log.Printf("Dumped system information to: %s", basePath+".tar.gz")
 	}()
 
+	log.Print("Starting RHMAP System Dump Tool...")
 	log.Println("Running tasks...")
 
-	// Run at most N tasks in parallel, and wait for all of them to
-	// complete.
+	tasks := GetAllTasks(basePath)
+	e := make(chan error)
```
rhcarvalho: I'll rename this to `results`, should make the intent clearer.

main.go
```diff
@@ -162,19 +168,6 @@ func main() {
 	log.SetOutput(io.MultiWriter(os.Stderr, logfile))
 	fileOnlyLogger := log.New(logfile, "", log.LstdFlags)
 
-	log.Print("Starting RHMAP System Dump Tool...")
-	log.Print("Preparing tasks...")
```
rhcarvalho: Now that tasks are "prepared" and run simultaneously, we don't need the "Preparing tasks" anymore, we go straight to "Running tasks".
We've introduced the "Preparing tasks" message when the time taken in that step started to be significant. And now it is gone :)

tasks.go
```diff
@@ -1,125 +1,121 @@
...
+		// Add tasks to fetch Nagios data.
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			GetNagiosTasks(projects, basepath, tasks)
+		}()
+
+		// Add check tasks.
+		wg.Add(1)
+		go func() {
```
philbrookes: Is there a way to guarantee that these check tasks will run after the dump tasks. Currently not a problem but when these rely on reading the files dumped in the dump tasks, they will fail if they run before the file has been dumped.

rhcarvalho: Good question. We can definitely arrange for them to run after the others.

Do you prefer we do it now, or just add a note in case that condition becomes true?

rhcarvalho: We can do it for any set of tasks by sending them after `wg.Wait()` ([line 108](https://github.com/feedhenry/fh-system-dump-tool/pull/23/files#diff-2628f4a06e8be3343e85abecc49fae35R108)).

main.go
```diff
@@ -33,9 +32,9 @@ const (
 )
 
 var (
-	maxParallelTasks = flag.Int("p", runtime.NumCPU(), "max number of tasks to run in parallel")
-	maxLogLines      = flag.Int("max-log-lines", defaultMaxLogLines, "max number of log lines fetched with oc logs")
-	printVersion     = flag.Bool("version", false, "print version and exit")
+	concurrentTasks = flag.Int("p", runtime.NumCPU(), "number of tasks to run concurrently")
```
rhcarvalho: "parallel" was misleading. One can call with "-p 1000" and it doesn't mean that there will be parallel execution of 1000 tasks, unless there are 1000+ CPU cores.

main.go
```diff
@@ -162,19 +161,6 @@ func main() {
...
-	log.Print("Starting RHMAP System Dump Tool...")
-	log.Print("Preparing tasks...")
-
-	tasks, err := GetAllTasks(basePath)
-	if err != nil {
-		printError(err)
-		defer os.Exit(1)
-	}
-	if len(tasks) == 0 {
-		log.Print("No tasks found to execute.")
```
rhcarvalho: @philbrookes now we don't handle this case anymore.

No tasks really mean we've made a programming mistake, don't you think?

rhcarvalho:
118 lines in, 118 lines out. I like that :smile:

main.go
philbrookes: yes, that's fine with me.

tasks.go
philbrookes: I'd think considering that once the refactor to writing files is completed the analysis tasks will be rewritten to read from the dumped files, we should make this change immediately.

rhcarvalho: Ok, will do it.

tasks.go
```diff
@@ -1,125 +1,162 @@
...
+			defer wg.Done()
+			GetNagiosTasks(tasks, projects, basepath)
+		}()
+
+		wg.Wait()
+
+		// After all other tasks are done, add analysis tasks. We want
+		// to run them strictly later so that they can leverage the
+		// output of commands executed previously by other tasks, e.g.,
+		// reading resource definitions.
```
rhcarvalho: @philbrookes this is what you asked for. Analysis tasks will run strictly after all others.

philbrookes:
:+1: LGTM

rhcarvalho:
Updated some code comments.

philbrookes:
When I run this locally, I get quite a lot of error output now:
https://gist.github.com/philbrookes/86e93b3cb956556708aef4ee73225552

I know this is coming from the log-previous as there is no previous logs to get, but it's not a great user experience, it looks like it's broken, when it is not.

rhcarvalho:
> When I run this locally, I get quite a lot of error output now:
> https://gist.github.com/philbrookes/86e93b3cb956556708aef4ee73225552
> 
> I know this is coming from the log-previous as there is no previous logs to get, but it's not a great user experience, it looks like it's broken, when it is not.

I agree!

I want to make the errors visible at first, and my idea for later was to filter things that are unlikely to represent a problem. The previous logs is one such thing.

My decision was based on the recent conversation we had about the "analysis" part outputting things that "may" be a problem, though we can't confirm it. So that users go have a look at things before reaching support.

philbrookes:
yep that's a good idea, noisy first, and then quieting the noise is a great approach

PR #24, opened 2016-08-16
output error if Nagios is missing

philbrookes:
# Motivation
The system dump tool is much more valuable when Nagios is present in the RHMAP projects. Currently if the Nagios pod is missing this is silently ignored. Now we will output a message to the user letting them know that the tool would be empowered were Nagios present.

# Changes
Check if any nagios pods were discovered, if none were found output a message to the user.

### DIFF
--- a/tasks.go
+++ b/tasks.go
@@ -93,7 +93,7 @@ func GetAllTasks(basepath string) <-chan Task {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			GetNagiosTasks(tasks, projects, basepath)
+			GetNagiosTasks(tasks, projects, basepath, getResourceNamesBySubstr)
 		}()
 
 		wg.Wait()
@@ -117,15 +117,18 @@ func NewError(err error) Task {
 }
 
 // GetNagiosTasks sends tasks to dump Nagios data for each project that contain
-// a Nagios pod.
-func GetNagiosTasks(tasks chan<- Task, projects []string, basepath string) {
+// a Nagios pod. This function will output an error to the user if no Nagios pods
+// were found in any projects.
+func GetNagiosTasks(tasks chan<- Task, projects []string, basepath string, resourceFactory ResourceMatchFactory) {
+	foundANagiosPod := false
 	for _, p := range projects {
-		pods, err := getResourceNamesBySubstr(p, "pod", "nagios")
+		pods, err := resourceFactory(p, "pod", "nagios")
 		if err != nil {
 			tasks <- NewError(err)
 			continue
 		}
 		for _, pod := range pods {
+			foundANagiosPod = true
 			outFor := outToFile(basepath, "dat", "nagios")
 			errOutFor := outToFile(basepath, "stderr", "nagios")
 			tasks <- GetNagiosStatusData(p, pod, outFor, errOutFor)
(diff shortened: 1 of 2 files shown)


tasks.go
```diff
@@ -135,6 +138,10 @@ func GetNagiosTasks(tasks chan<- Task, projects []string, basepath string) {
 			tasks <- GetNagiosHistoricalData(p, pod, outFor, errOutFor)
 		}
 	}
+
+	if !foundANagiosPod {
+		fmt.Println("We could not find a Nagios pod in any project, which will reduce the usefulness of this tool, for a full analysis please ensure that Nagios is running in all RHMAP projects.")
```
david-martin: Looks good.
Some minor rephrasing might make it read better though
e.g. "A Nagios pod could not be found in any project. For a more thorough analysis, please ensure Nagios is running in all RHMAP projects"

tasks.go
```diff
@@ -135,6 +138,10 @@ func GetNagiosTasks(tasks chan<- Task, projects []string, basepath string) {
 			tasks <- GetNagiosHistoricalData(p, pod, outFor, errOutFor)
 		}
 	}
+
+	if !foundANagiosPod {
+		fmt.Println("A Nagios pod could not be found in any project. For a more thorough analysis, please ensure Nagios is running in all RHMAP projects.")
```
rhcarvalho: We should not print to stdout, or it won't be included in the `dump.log`.
Would it make sense to make it a task error?

tasks.go
```diff
@@ -117,15 +117,18 @@ func NewError(err error) Task {
 }
 
 // GetNagiosTasks sends tasks to dump Nagios data for each project that contain
-// a Nagios pod.
+// a Nagios pod. This function will output an error to the user if no Nagios pods
+// were found in any projects.
```
rhcarvalho: Would be nicer to see this in a test case, we've got nothing for Nagios.

rhcarvalho:
LGTM

nagios_test.go
```diff
@@ -0,0 +1,29 @@
...
+	})
+	close(tasks)
+
+	expected := "A Nagios pod could not be found in any project. For a more thorough analysis, please ensure Nagios is running in all RHMAP projects."
+
+	for task := range tasks {
+		if task() == nil {
+			t.Fatal("task(): got nil, Expected Error")
+		}
+		if task().Error() != expected {
```
rhcarvalho: This executes the task for a second time.

tasks.go
```diff
@@ -117,15 +117,18 @@ func NewError(err error) Task {
...
 	for _, p := range projects {
-		pods, err := getResourceNamesBySubstr(p, "pod", "nagios")
+		pods, err := resourceFactory(p, "pod", "nagios")
 		if err != nil {
 			tasks <- NewError(err)
 			continue
 		}
 		for _, pod := range pods {
+			foundANagiosPod = true
 			outFor := outToFile(basepath, "dat", "nagios")
```
philbrookes: I don't see an easy way to test the inside of this for loop without passing the outToFile function so it can be mocked in the test. Is it worth doing that now?

nagios_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"testing"
+)
+
+
+func TestGetNagiosTasksErrorWithMissingNagiosPod(t *testing.T) {
+	tasks := make(chan Task, 1)
+	GetNagiosTasks(tasks, []string{"project1"}, "basepath", func(project, resource, substr string) ([]string, error) {
+		return []string{}, nil
+	})
+	close(tasks)
```
rhcarvalho: This block is not properly synchronized. There's no guarantee here that `GetNagiosTasks` will send any task before the channel is closed, and sending on a closed channel panics.

rhcarvalho:
@philbrookes I'm okay with deferring tests for later, as I'm refactoring big chunks. The more code here, the more conflicts I'll have to solve :P

tasks.go
rhcarvalho: No.

nagios_test.go
philbrookes: I'm not sure I follow you. The test and the GetNagiosTasks are both synchronous code on the same thread, how could the channel be written to after it is closed here?

rhcarvalho: You are right, sorry.

PR #25, opened 2016-08-16
Rhmap 9066

philbrookes:
# Motivation
Currently the dump created by the system dump tool does not include the output from`oc adm diagnostics` this change adds the output to the dump.

# Changes
- Add a new task to run `oc adm diagnostics` on the platform and add it to a `diagnostics` folder at the root level of the dump.
- Change the `OutToFile` to `OutToFileForProject` and add a new `OutToFileForPlatform` which writes to the root directory of the dumped archive.

### DIFF
--- a/dump_writers.go
+++ b/dump_writers.go
@@ -10,12 +10,34 @@ import (
 // particular resource type within a project.
 type projectResourceWriterCloserFactory func(project, resource string) (io.Writer, io.Closer, error)
 
-// outToFile returns a function that creates an io.Writer that writes to a file
+// A platformResourceWriterFactory generates io.Writers for dumping data of a
+// particular resource type.
+type platformResourceWriterCloserFactory func(resource string) (io.Writer, io.Closer, error)
+
+// outToFileForProject returns a function that creates an io.Writer that writes to a file
 // in basepath with extension, given a project and resource. The scope can be passed as a value to group the
 // files under, or assigned an empty value to have the files grouped directly under the project.
-func outToFile(basepath, extension, scope string) projectResourceWriterCloserFactory {
+func outToFileForProject(basepath, extension, scope string) projectResourceWriterCloserFactory {
 	return func(project, resource string) (io.Writer, io.Closer, error) {
 		scopePath := filepath.Join(basepath, "projects", project, scope)
+		err := os.MkdirAll(scopePath, 0770)
+		if err != nil {
+			return nil, nil, err
+		}
+		f, err := os.Create(filepath.Join(scopePath, resource+"."+extension))
+		if err != nil {
+			return nil, nil, err
+		}
+		return f, f, nil
+	}
+}
+
+// outToFileForPlatform returns a function that creates an io.Writer that writes to a file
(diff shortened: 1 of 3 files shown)


oadm.go
```diff
@@ -0,0 +1,45 @@
...
+	}, outFor, errOutFor)
+}
+
+// A getOcAdmDiagnosticsCmdFactory generates commands to get the raw oc admin diagnostics
+// data.
+type getOcAdmDiagnosticsCmdFactory func() *exec.Cmd
+
+// oadmData will dump the output of 'oc admin diagnostics' into the stdout writer any stderr
+// output is sent to the stderr writer
+func oadmData(cmdFactory getOcAdmDiagnosticsCmdFactory, outFor, errOutFor platformResourceWriterCloserFactory) Task {
```
philbrookes: I feel like this function is happening frequently. I will wait for the refactor from @rhcarvalho as I think it will probably address this, but if not. I will try to reduce the amount of duplicated code.

rhcarvalho: Using `Runner` from #26, this PR should be a lot less code :)

oadm.go
```diff
@@ -0,0 +1,45 @@
+package main
+
+import (
+	"os/exec"
+)
+
+// OadmData is a task factory for tasks that fetch the output of oc admin diagnostics.
```
rhcarvalho: Isn't it "oadm diagnostics"?

I didn't know, but these two seem to produce the same output:

```
$ oc adm diagnostics
$ oadm diagnostics
```

I don't know what the openshift-client package on RHEL installs, but I think it would be reasonable to use the `oc adm` version so that we have a binary dependency on a single name.

If we use `oadm`, we should document that new dependency.

oadm.go
```diff
@@ -0,0 +1,45 @@
...
+// The task uses outFor and errOutFor to get io.Writers to write, respectively, the
+// stdout output and stderr output.
+func OadmData(outFor, errOutFor platformResourceWriterCloserFactory) Task {
+	return oadmData(func() *exec.Cmd {
+		return exec.Command("oc", "adm", "diagnostics")
```
rhcarvalho: Oh yeah, +1 for this instead of `oadm`.

dump_writers.go
```diff
@@ -27,3 +31,21 @@ func outToFile(basepath, extension, scope string) projectResourceWriterCloserFac
 		return f, f, nil
 	}
 }
+
+// outToFileForPlatform returns a function that creates an io.Writer that writes to a file
+// in basepath with extension, given a project and resource. The scope can be passed as a value to group the
+// files under, or assigned an empty value to have the files grouped directly under the project.
+func outToFileForPlatform(basepath, extension, scope string) platformResourceWriterCloserFactory {
```
rhcarvalho: Using the `Runner` from #26 we don't need this.

This PR will be essentially:

```go
cmd := exec.Command("oc", "adm", "diagnostics")
path := "oc_adm_diagnostics"
runner.Run(cmd, path)
```

PR #26, opened 2016-08-16
Change abstraction for saving output of commands

rhcarvalho:
The `outFor` and `errOutFor` abstractions are getting harder and harder to use as we add more code.
Tests were brittle, near meaningless, and we were writing a lot of boilerplate code to get things done.

This PR introduces a `DumpRunner`. It can run commands and save output to files.
It will create a "name.stderr" file with stderr from commands, if any.

It should be now easier to write tests and code.

This PR shows how the logs tasks gets refactored to use the new abstraction. Future PRs will refactor the rest, but I decided not to submit all at once to keep this minimally reviewable.

-------

https://issues.jboss.org/browse/RHMAP-9533

### DIFF
--- a/runner.go
+++ b/runner.go
@@ -1,0 +1,109 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"sync"
+)
+
+// A Runner runs commands.
+type Runner interface {
+	Run(cmd *exec.Cmd, path string) error
+}
+
+// DumpRunner is a Runner that dumps command execution output to disk.
+type DumpRunner struct {
+	dir string
+}
+
+var _ Runner = (*DumpRunner)(nil)
+
+// NewDumpRunner creates a DumpRunner.
+func NewDumpRunner(dir string) *DumpRunner {
+	return &DumpRunner{
+		dir: dir,
+	}
+}
+
+// Run runs cmd and saves the stdout to path, relative to r.dir. Stderr, if any,
+// goes to path.stderr. Parent directories are created if necessary. Path cannot
+// be empty.
+func (r *DumpRunner) Run(cmd *exec.Cmd, path string) error {
+	if path == "" {
+		return fmt.Errorf("command %q: missing path to output", strings.Join(cmd.Args, " "))
+	}
+
+	basedir := filepath.Join(r.dir, filepath.Dir(path))
+	if err := os.MkdirAll(basedir, 0770); err != nil {
+		return err
+	}
+
+	stdout, err := os.Create(filepath.Join(r.dir, path))
+	if err != nil {
+		return err
+	}
+	defer stdout.Close()
+
+	var pathStderr = path + ".stderr"
+	stderr := &lazyFileWriter{path: filepath.Join(r.dir, pathStderr)}
+	defer stderr.Close()
+
+	cmd.Stdout = io.MultiWriter(filterWriters(cmd.Stdout, stdout)...)
+	cmd.Stderr = io.MultiWriter(filterWriters(cmd.Stderr, stderr)...)
+
+	if err := cmd.Run(); err != nil {
+		var b []byte
+		if stderr.file != nil {
+			stderr.file.Seek(0, os.SEEK_SET)
+			b, _ = ioutil.ReadAll(stderr.file)
+		}
+		return fmt.Errorf("command %q: %s: %s", strings.Join(cmd.Args, " "), err, b)
+	}
+	return nil
+}
+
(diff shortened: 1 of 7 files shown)


runner.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return err
+	}
+	defer stdout.Close()
+
+	var pathStderr = path + ".stderr"
+	stderr := &lazyFileWriter{path: filepath.Join(r.dir, pathStderr)}
+	defer stderr.Close()
+
+	cmd.Stdout = io.MultiWriter(filterWriters(cmd.Stdout, stdout)...)
```
philbrookes: Does this mean that in the case of Nagios, we will still be writing to a tar file for now?

runner_test.go
```diff
@@ -0,0 +1,151 @@
...
+	names := []string{"out"}
+	contents := []string{cmd.Args[len(cmd.Args)-1] + "\n"}
+
+	if err := dr.Run(cmd, names[len(names)-1]); err != nil {
+		t.Errorf("Run(%q) = %v, want %v", cmd.Args[3:], err, nil)
+	}
+	if err := dirHasExactFiles(dir, names, contents); err != nil {
+		t.Error(err)
+	}
+	if got, want := b.String(), contents[len(contents)-1]; got != want {
```
philbrookes: this is cool, I didn't know you could do multiple comparisons like this, I assume this is compared with `and`? In other words, it's the same as writing this:
```
if got == b.String() && wants == contents[len(contents)-1]{
```

philbrookes: oh no, I see, I have misunderstood this. This is more like doing:
```
got := b.String()
want := contents[len(contents)-1]
if got == want {
```

runner_test.go
```diff
@@ -0,0 +1,151 @@
...
+	cmd := helperCommand("echo", "ok")
+
+	// Set cmd.Stdout to later ensure that dr.Run will write to both b and a
+	// file.
+	var b bytes.Buffer
+	cmd.Stdout = &b
+
+	// These are the files and contents we expect to see:
+	names := []string{"out"}
+	contents := []string{cmd.Args[len(cmd.Args)-1] + "\n"}
```
philbrookes: would this be easier as: `contents := []string{cmd.Args[1:]}`?

tasks.go
```diff
@@ -13,8 +13,8 @@ type Task func() error
 
 // RunAllTasks runs all tasks known to the dump tool using concurrent workers.
 // Dump output goes to path.
-func RunAllTasks(path string, workers int) {
-	tasks := GetAllTasks(path)
+func RunAllTasks(runner Runner, path string, workers int) {
```
philbrookes: why is `path string` required here, considering that the runner is constructed with it here: https://github.com/feedhenry/fh-system-dump-tool/pull/26/files#diff-7ddfb3e035b42cd70649cc33393fe32cR191

runner.go
rhcarvalho: Yes, this just writes the tar file. However, we could add an option to the runner to automatically untar when desired?
What analysis will we be doing on the tar contents? I'd defer implementing that a bit more until we know what we will be reading from those files.

runner_test.go
rhcarvalho: Yes, as you said above.

The [`if` statement](https://golang.org/ref/spec#If_statements) takes an optional statement, in this case used for doing multiple assignment (or "[tuple assignment](https://golang.org/ref/spec#Assignments)").

@philbrookes the difference to the code you wrote above is that `got` and `want` exist only in the scope of the `if`.

runner.go
philbrookes: that's fine, I just wanted to ensure I was understanding this properly. For now, staying as a tar is fine. I don't think any of the planned automatic checks will need this data.

runner_test.go
rhcarvalho: Once I had this hard-coded, perhaps it was easier to read...

```go
	contents := []string{"ok\n"}
```

> would this be easier as: `contents := []string{cmd.Args[1:]}`?

That wouldn't compile, as it is attempting to put a slice where we expect a string.

And `cmd.Args[1:]` is not `[]string{"ok"}` as one may imagine, because `helperCommand` returns a `*exec.Cmd` that runs the test binary itself with crafted arguments, among those `"echo"` and `"ok"`.

We also need to include the newline character written to stdout.

tasks.go
rhcarvalho: Well observed.

This is temporary, we can't remove it yet because definitions, nagios, and analysis checks still depend on it.
It gets removed with the future commits that refactor the other parts.

philbrookes:
:+1: LGTM

PR #28, opened 2016-08-17
Change definitions and Nagios tasks to use Runner

rhcarvalho:
This follows up on #26, changing more parts of the code base to use a `Runner` instead of the `outFor` and `errOutFor` abstractions.

There is one missing part before we can get rid of old code: the analysis tasks. They are notably more complicated, will go as yet another follow up.

-----------
https://issues.jboss.org/browse/RHMAP-9535

### DIFF
--- a/definitions.go
+++ b/definitions.go
@@ -1,6 +1,26 @@
 package main
 
-import "os/exec"
+import (
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+// ResourceDefinition is a task factory for tasks that fetch the JSON resource
+// definition for the given resource in the given project.
+func ResourceDefinition(r Runner, project, resource string) Task {
+	return func() error {
+		// NOTE: we could fetch all resources of all types in a single
+		// call to oc, by passing a comma-separated list of resource
+		// types. Instead, we call oc multiple times to send the output
+		// to different files without processing the contents of the
+		// output from oc.
+		cmd := exec.Command("oc", "-n", project, "get", resource, "-o=json")
+		fname := strings.Replace(resource, "/", "_", -1) + ".json"
+		path := filepath.Join("projects", project, "definitions", fname)
+		return r.Run(cmd, path)
+	}
+}
 
 // ResourceDefinitions is a task factory for tasks that fetch the JSON resource
 // definition for all given types in project. For each resource type, the task
--- a/nagios.go
+++ b/nagios.go
@@ -1,55 +1,28 @@
 package main
 
 import (
-	"fmt"
 	"os/exec"
+	"path/filepath"
 )
 
-// GetNagiosStatusData is a task factory for tasks that fetch a JSON representation of
-// the provided project. The task uses outFor and errOutFor to get io.Writers
-// to write, respectively, the JSON output and any eventual error message.
-func GetNagiosStatusData(project, pod string, outFor, errOutFor projectResourceWriterCloserFactory) Task {
-	return retrieveNagiosData(func() *exec.Cmd {
-		return exec.Command("oc", "exec", pod, "--", "cat", "/var/log/nagios/status.dat")
-	}, project, pod, "status", outFor, errOutFor)
-}
-
-// GetNagiosHistoricalData is a task factory for tasks that fetch a JSON representation of
-// the provided project. The task uses outFor and errOutFor to get io.Writers
-// to write, respectively, the JSON output and any eventual error message.
-func GetNagiosHistoricalData(project, pod string, outFor, errOutFor projectResourceWriterCloserFactory) Task {
-	return retrieveNagiosData(func() *exec.Cmd {
-		return exec.Command("oc", "exec", pod, "--", "tar", "-c", "-C", "/var/log/nagios", "archives")
-	}, project, pod, "history", outFor, errOutFor)
-}
-
-// A getNagiosDataCmdFactory generates commands to get the raw nagios status data
-// in the provided project from the provided pod
-type getNagiosDataCmdFactory func() *exec.Cmd
-
-// retrieveNagiosData will dump a JSON representation of the current nagios status into
-// the io.Writer returned from outFor and log any errors it encounters doing so
-// into the io.Writer returned from errOutFor.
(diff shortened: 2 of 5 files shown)


logs.go
```diff
@@ -23,6 +23,53 @@ type LoggableResource struct {
...
+	)
+	for _, p := range projects {
+		for _, rtype := range resources {
+			names, err := GetResourceNames(p, rtype)
+			if err != nil {
+				errors = append(errors, err)
+				continue
+			}
+			for _, name := range names {
+				resources, err := GetLoggableResources(p, rtype, name)
```
philbrookes: Could we change this function name to something else? I initially thought this was a recursive function.

philbrookes:
:+1: This looks good to me, other than the initial confusion over the function name.

logs.go
rhcarvalho: Yes, I noticed the names (with a typo) when I moved stuff to the same file.
Now looks like a good PR to rename it, thanks!

tasks.go
```diff
@@ -46,6 +53,11 @@ func RunAllTasks(runner Runner, path string, workers int) {
 		fmt.Fprint(os.Stderr, ".")
 	}
 	fmt.Fprintln(os.Stderr)
+
+	delta := time.Since(start)
+	// Remove sub-second precision.
+	delta -= delta % time.Second
+	log.Printf("Run %d tasks in %v.", taskCount, delta)
```
rhcarvalho: @philbrookes WDYT? I'm often curious to know tasks don't simply "disappear", or whether we are growing the number of tasks over time.

Worth noting that the number of tasks depend on the data in the cluster (e.g., more pods ￫ more tasks to fetch logs).

philbrookes: yeah, I think this is a great idea.

philbrookes: We could also add a number of projects here, to help with making sure no tasks are missing.

rhcarvalho:
@philbrookes scope reduced to the original intent.

definitions.go
```diff
@@ -1,6 +1,25 @@
...
+		// types. Instead, we call oc multiple times to send the output
+		// to different files without processing the contents of the
+		// output from oc.
+		cmd := exec.Command("oc", "-n", project, "get", resource, "-o=json")
+		path := filepath.Join("projects", project, "definitions", strings.Replace(resource, "/", "_", -1)+".json")
+		return r.Run(cmd, path)
+	}
+}
 
 // ResourceDefinitions is a task factory for tasks that fetch the JSON resource
```
rhcarvalho: Note that I had to leave some old code here, because the analysis tasks depend on it.

philbrookes:
:+1: cool, this looks good.

definitions.go
```diff
@@ -1,6 +1,26 @@
...
+func ResourceDefinition(r Runner, project, resource string) Task {
+	return func() error {
+		// NOTE: we could fetch all resources of all types in a single
+		// call to oc, by passing a comma-separated list of resource
+		// types. Instead, we call oc multiple times to send the output
+		// to different files without processing the contents of the
+		// output from oc.
+		cmd := exec.Command("oc", "-n", project, "get", resource, "-o=json")
+		fname := strings.Replace(resource, "/", "_", -1) + ".json"
+		path := filepath.Join("projects", project, "definitions", fname)
```
rhcarvalho: The path where we write the files should not have changed.
E.g.:

- `projects/rhmap-core/definitions/pods.json`
- `projects/rhmap-core/definitions/pod_supercore-kj34sd.json`

nagios.go
```diff
@@ -1,55 +1,28 @@
...
-	return retrieveNagiosData(func() *exec.Cmd {
-		return exec.Command("oc", "exec", pod, "--", "tar", "-c", "-C", "/var/log/nagios", "archives")
-	}, project, pod, "history", outFor, errOutFor)
+// GetNagiosStatusData is a task factory for tasks that fetch Nagios status from
+// the given pod in project.
+func GetNagiosStatusData(r Runner, project, pod string) Task {
+	return func() error {
+		cmd := exec.Command("oc", "exec", pod, "--", "cat", "/var/log/nagios/status.dat")
+		fname := pod + "_status.dat"
+		path := filepath.Join("projects", project, "nagios", fname)
```
rhcarvalho: Also here the path should not have changed.

nagios.go
```diff
@@ -1,55 +1,28 @@
...
-		cmd := cmdFactory()
-
-		if err := runCmdCaptureOutput(cmd, stdout, stderr); err != nil {
-			fmt.Fprint(stderr, err)
-			return err
-		}
-		return nil
+		cmd := exec.Command("oc", "exec", pod, "--", "tar", "-c", "-C", "/var/log/nagios", "archives")
+		fname := pod + "_history.tar"
+		path := filepath.Join("projects", project, "nagios", fname)
```
rhcarvalho: And finally here the path should not have changed.

Now paths are fully constructed in a single place, which I consider better than before. However there's still the problem I mentioned on IRC, manually writing these paths is error prone, we could get into a situation that a file name clashes with an existing directory or vice-versa.

tasks.go
```diff
@@ -116,28 +116,20 @@ func NewError(err error) Task {
...
 // GetNagiosTasks sends tasks to dump Nagios data for each project that contain
-// a Nagios pod. This function will output an error to the user if no Nagios pods
-// were found in any projects.
-func GetNagiosTasks(tasks chan<- Task, projects []string, basepath string, resourceFactory ResourceMatchFactory) {
+// a Nagios pod. It is an error if no projects contain a Nagios pod.
+func GetNagiosTasks(tasks chan<- Task, runner Runner, projects []string) {
 	foundANagiosPod := false
 	for _, p := range projects {
-		pods, err := resourceFactory(p, "pod", "nagios")
+		pods, err := getResourceNamesBySubstr(p, "pod", "nagios")
```
philbrookes: why is this no longer injected? Doesn't this make it harder to test in the future?

rhcarvalho: Since it is not tested, I went with simplifying the code.
Eventually `getResourceNamesBySubstr` should take the runner as argument.

Then using a fake runner we can test the whole thing without having to inject a lot of different functions.

PR #29, opened 2016-08-17
add oc adm diagnostics output to dump

philbrookes:
# Motivation
Currently the dump created by the system dump tool does not include the output from`oc adm diagnostics` this change adds the output to the dump.

# Changes
- Add a new task to run `oc adm diagnostics` on the platform and add it to the root level of the dump.

### DIFF
--- a/tasks.go
+++ b/tasks.go
@@ -96,6 +96,12 @@ func GetAllTasks(runner Runner, basepath string) <-chan Task {
 			GetNagiosTasks(tasks, projects, basepath, getResourceNamesBySubstr)
 		}()
 
+		wg.Add(1)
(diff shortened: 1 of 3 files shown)


tasks.go
```diff
@@ -118,6 +125,16 @@ func NewError(err error) Task {
 
 type ResourceMatchFactory func(project, resource, substr string) ([]string, error)
 
+
+// GetOcAdmDiagnosticsTasks sends tasks to fetch the oc adm diagnostics result
```
rhcarvalho: End the comment with a final stop "." and LGTM.

rhcarvalho:
LGTM

PR #30, opened 2016-08-17
Fetch definitions of PVs and PVCs

rhcarvalho:
These are useful to understand the state of storage objects in the cluster.

https://issues.jboss.org/browse/RHMAP-9543

### DIFF
--- a/definitions.go
+++ b/definitions.go
@@ -9,6 +9,10 @@ import (
 // GetResourceDefinitionTasks sends tasks to fetch the definitions of all
 // resources in all projects.
 func GetResourceDefinitionTasks(tasks chan<- Task, runner Runner, projects, resources []string) {
+	// NOTE: we could fetch all resources of all types in a single call to
+	// oc, by passing a comma-separated list of resource types. Instead, we
+	// call oc multiple times to send the output to different files without
+	// processing the contents of the output from oc.
 	for _, p := range projects {
 		for _, resource := range resources {
 			tasks <- ResourceDefinition(runner, p, resource)
@@ -20,11 +24,6 @@ func GetResourceDefinitionTasks(tasks chan<- Task, runner Runner, projects, reso
 // definition for the given resource in the given project.
 func ResourceDefinition(r Runner, project, resource string) Task {
 	return func() error {
-		// NOTE: we could fetch all resources of all types in a single
-		// call to oc, by passing a comma-separated list of resource
-		// types. Instead, we call oc multiple times to send the output
-		// to different files without processing the contents of the
-		// output from oc.
 		var args []string
 		if project != "" {
 			args = append(args, "-n", project)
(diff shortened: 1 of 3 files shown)


philbrookes:
:+1: This looks good to me.

tasks.go
```diff
@@ -54,7 +54,10 @@ func RunAllTasks(runner Runner, path string, workers int) {
 // FIXME: GetAllTasks should not need to know about basepath.
 func GetAllTasks(runner Runner, basepath string) <-chan Task {
 	var (
-		resources = []string{"deploymentconfigs", "pods", "services", "events"}
+		resources = []string{
+			"deploymentconfigs", "pods", "services", "events",
+			"persistentvolumes", "persistentvolumeclaims",
```
rhcarvalho: As @david-martin noted, PVs are cluster scoped, so we should fetch it only once per run of the tool, not once per project.

philbrookes: yeah, I overlooked that alright! Similar to the diagnostics task in that case then.

david-martin:
It might be better to get persistentvolumes once per dump as it currently isn't possible to use `-n` to restrict them to a project

rhcarvalho:
Got this working, but based the work on top of code from currently in-flight PRs. Will wait for them to merge first to update this with only the relevant patch.

rhcarvalho:
Updated implementation to get every PVC of every project + all PVs of the cluster.

definitions.go
```diff
@@ -20,14 +24,21 @@ func GetResourceDefinitionTasks(tasks chan<- Task, runner Runner, projects, reso
 // definition for the given resource in the given project.
 func ResourceDefinition(r Runner, project, resource string) Task {
```
philbrookes: as project can be empty, does it make more sense to make it the last argument?
i.e. 
```
ResourceDefinition(runner, "persistentvolumes", "")
```
Rather than:
```
ResourceDefinition(runner, "", "persistentvolumes")
```

or possibly we could wrap the method with something like:
```
func PlatformResourceDefinition(r Runner, resource string) Task {
    return ResourceDefinition(r, resource, "")
}
```
What do you think?

rhcarvalho: Technically `resource` could be the empty string too.
I prefer the "logical" order that one normally specify from broader to narrower: project -> resource type -> resource name, project -> pod -> pod name -> container name, ...

I don't see much value in wrapping with another function. One can still pass any resource, even if it is not "cluster scoped".

E.g. calling `PlatformResourceDefinition(r, "services")` will return all services in the current project -- but the "current project" is out of control of the dump tool, I'd say that would be a mistake to make such call in the context of the tool.

In the end of the day `ResourceDefinition` is just a wrapper to `oc [-n PROJ] get [KIND/NAME] [...]`, it could be called `OcGet` :D

philbrookes: Sure OK, that makes sense.

philbrookes:
:+1: LGTM

PR #31, opened 2016-08-17
dump millicore properties - RHMAP-8123

philbrookes:
# Motivation 
The dump tool should contain the millicore properties file.

# Changes
- Add tasks to dump millicore properties file.
- Add tests to test the new millicore.go code.

### DIFF
--- a/millicore.go
+++ b/millicore.go
@@ -1,0 +1,27 @@
+package main
+
+import (
+	"os/exec"
+	"path/filepath"
+)
+
+func GetMillicoreConfigTasks(tasks chan<- Task, runner Runner, projects []string, resourceFactory ResourceMatchFactory) {
+	for _, p := range projects {
+		pods, err := resourceFactory(p, "pod", "millicore")
+		if err != nil {
+			tasks <- NewError(err)
+			continue
+		}
+		for _, pod := range pods {
+			tasks <- GetMillicoreConfig(runner, p, pod)
+		}
+	}
+}
+
+func GetMillicoreConfig(r Runner, project, pod string) Task {
+	return func() error {
+		cmd := exec.Command("oc", "exec", pod, "--", "cat", "/etc/feedhenry/cluster-override.properties")
+		path := filepath.Join("projects", project, "millicore", pod+"_cluster-override.properties")
+		return r.Run(cmd, path)
+	}
+}
--- a/millicore_test.go
+++ b/millicore_test.go
@@ -1,0 +1,49 @@
+package main
+
+import (
+	"errors"
+	"path/filepath"
+	"reflect"
+	"testing"
+)
+
+func TestGetMillicoreConfigTasks(t *testing.T) {
+	tasks := make(chan Task, 1)
+	runner := &FakeRunner{}
+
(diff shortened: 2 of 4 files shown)


millicore.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"os/exec"
+	"path/filepath"
+)
+
+func GetMillicoreConfigTasks(tasks chan<- Task, runner Runner, projects []string, resourceFactory ResourceMatchFactory) {
+	for _, p := range projects {
+		pods, err := resourceFactory(p, "pod", "millicore")
+		if err != nil {
+			tasks <- NewError(err)
```
rhcarvalho: Will this produce an error when a millicore pod is not found?

millicore.go
```diff
@@ -0,0 +1,28 @@
...
+		}
+	}
+}
+
+func GetMillicoreConfig(r Runner, project, pod string) Task {
+	return func() error {
+		cmd := exec.Command("oc", "exec", pod, "--", "cat", "/etc/feedhenry/cluster-override.properties")
+		path := filepath.Join("projects", project, "millicore", pod+"_cluster-override.properties")
+		return r.Run(cmd, path)
+		return nil
```
rhcarvalho: unreachable code

millicore_test.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+}
+
+func TestMillicorePodError(t *testing.T) {
+	tasks := make(chan Task, 1)
+	runner := &FakeRunner{}
+
+	GetMillicoreConfigTasks(tasks, runner, []string{"project1"}, func(project, resource, substr string) ([]string, error) {
+		return nil, errors.New("error retrieving pods")
```
rhcarvalho: This seems strange to me. Why do we test this, if the real thing we pass is `getResourceNamesBySubstr` and it doesn't have that behavior?

millicore.go
philbrookes: no, added a check for empty result set, to send an error into the Tasks channel.

millicore_test.go
philbrookes: The actualy `getResourceNamesBySubstr` can return an error when the oc command returns an error. This test is just testing that when an error is returned from that function it does come through the channel correctly. 

It also covers an otherwise uncovered part of the millicore code.

millicore.go
rhcarvalho: I don't think that should be an error... we don't expect every project to have a "millicore" pod. Similar story with Nagios.

philbrookes: yeah OK, that makes sense thinking about it, as Nagios will be showing quite a lot of errors if millicore is missing where it is expected.

rhcarvalho:
LGTM

pmccarthy:
Ran this locally, as it turns out I needed to upgrade to to OC v3.2 as the 'oc adm diagnostics' is not available on v3.1 - something to note. 

Once I had upgraded, all seems to have worked ok, had a millicore dump file with all expected properties and values 👍

rhcarvalho:
@pmccarthy the dump tool should have worked and should have produced the millicore information even when `oc adm diagnostics` is not available. Can you confirm whether that was the case?

PR #32, opened 2016-08-18
Reorganize code

rhcarvalho:
This moves code away from tasks.go to more specific files.

It should make it easier to find all code related to a certain task type.


nagios.go
```diff
@@ -26,3 +49,22 @@ func GetNagiosHistoricalData(r Runner, project, pod string) Task {
 		return r.Run(cmd, path)
 	}
 }
+
+// getResourceNamesBySubstr returns a list of names for the provided resource type that contain
+// the provided string, in the provided project.
+func getResourceNamesBySubstr(project, resource, substr string) ([]string, error) {
```
philbrookes: not sure this should be here, it will be needed by the millicore tasks too.

rhcarvalho: Hmmm, for sure it had nothing to do with logs. Where should it go then?

philbrookes: maybe we need a new file for shared functions and that? Like `shared.go` or something?

philbrookes:
:+1: LGTM

nagios.go
rhcarvalho: Talked on IRC, staying here for now, may move somewhere else in the future when it will have more callers.

PR #33, opened 2016-08-18
Cleanup logs tasks

rhcarvalho:
Make each task map to a call to 'oc logs', so that we can have more fine
grained tasks that can run concurrently.

Eventually also fixes a problem of having similarly named functions
GetLogabbleResources and GetLoggableResources, including a typo.


philbrookes:
:+1: LGTM, I like the new fakeRunner, looks useful

PR #34, opened 2016-08-18
Add tasks to get cluster metadata

rhcarvalho:
WIP note: implementation works, but we might want to put things into
functions, separate files, etc.

----------

https://issues.jboss.org/browse/RHMAP-9091
https://issues.jboss.org/browse/RHMAP-9536

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,6 +7,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


philbrookes:
I think we should have the version of the dump tool in this metadata too.

rhcarvalho:
> I think we should have the version of the dump tool in this metadata too.

Yes, it is in the works.

philbrookes:
:+1: LGTM

rhcarvalho:
Moved code fetching openshift metadata to "openshift_metadata.go"

PR #36, opened 2016-08-22
refactor analysis to read from dump

philbrookes:
This is a WIP, opening the PR to allow for feedback on current progress.

# Motivation
Analysis tasks currently pull data from the platform. This rewrite will make them run on the dumped data instead.

Analysis tasks currently could potentially run while dump tasks are still executing, this rewrite ensures that they will run after all dump tasks are completed.

# Changes
Update analysis tasks and tests to read from the dumped files

Update tasks.go to run analysis tasks  after all the dump tasks are completed.

### DIFF
--- a/analysis.go
+++ b/analysis.go
@@ -3,9 +3,8 @@ package main
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
-	"io"
 	"io/ioutil"
+	"path/filepath"
 	"strings"
 )
 
@@ -50,60 +49,51 @@ type DeploymentConfigs struct {
 	} `json:"items"`
 }
 
-type CheckTask func(string, io.Writer) (Result, error)
+type CheckTask func(string, string) (Result, error)
+
+func GetAnalysisTasks(tasks chan<- Task, basepath string, projects []string, results chan<- CheckResults) {
+	// Platform-wide analysis goes here
+
+	// project specific analysis in here
+	for _, p := range projects {
+		tasks <- CheckProjectTask(p, basepath, results)
+	}
+}
 
 // CheckTasks is a task factory for tasks that diagnose system conditions.
-func CheckTasks(project string, outFor, errOutFor projectResourceWriterCloserFactory) Task {
-	return checkTasks(func() []CheckTask {
+func CheckProjectTask(project, basepath string, results chan<- CheckResults) Task {
+	return checkProjectTask(func() []CheckTask {
 		return []CheckTask{CheckImagePullBackOff, CheckDeployConfigsReplicasNotZero}
-	}, project, outFor, errOutFor)
+	}, project, basepath, results)
 }
 
 // A getProjectCheckFactory generates tasks to diagnose system conditions.
 type getProjectCheckFactory func() []CheckTask
 
 type CheckResults struct {
+	Scope   string `json:"scope"`
 	Results []Result
 }
 
 // checkTasks executes all the CheckTasks returned from the supplied
 // checkFactory against the specified project. The results of the checks are
-// combined into a single JSON object and written to the writer returned from
-// outFor, any errors that occur during the test are written to the writer
-// returned from errOutFor.
-func checkTasks(checkFactory getProjectCheckFactory, project string, outFor, errOutFor projectResourceWriterCloserFactory) Task {
+// combined into a single JSON object and returned
+func checkProjectTask(checkFactory getProjectCheckFactory, project, basepath string, results chan<- CheckResults) Task {
 	return func() error {
-		stdOut, stdOutCloser, err := outFor(project, "analysis")
-		if err != nil {
-			return err
-		}
-		defer stdOutCloser.Close()
-		stdErr, stdErrCloser, err := errOutFor(project, "analysis")
-		if err != nil {
-			return err
-		}
-		defer stdErrCloser.Close()
-
-		results := CheckResults{Results: []Result{}}
-
+		result := CheckResults{Scope: project, Results: []Result{}}
 		checks := checkFactory()
 
 		var errors errorList
 		for _, check := range checks {
-			res, err := check(project, stdErr)
+			res, err := check(project, basepath)
 			if err != nil {
 				errors = append(errors, err)
 			}
-			results.Results = append(results.Results, res)
+			result.Results = append(result.Results, res)
 
 		}
 
-		output, err := json.MarshalIndent(results, "", "    ")
-		if err != nil {
-			errors = append(errors, err)
-		}
-
-		stdOut.Write(output)
+		results <- result
 
 		if len(errors) > 0 {
 			return errors
@@ -112,33 +102,15 @@ func checkTasks(checkFactory getProjectCheckFactory, project string, outFor, err
 	}
 }
 
-// getResourceStruct retrieves the requested resource in the supplied project
-// from the platform and parse the JSON into the supplied interface.
-func getResourceStruct(project, resource string, dest interface{}) error {
-	stdOut := bytes.NewBuffer([]byte{})
-	stdErr := bytes.NewBuffer([]byte{})
-	outFor := func(project, resource string) (io.Writer, io.Closer, error) {
-		return stdOut, ioutil.NopCloser(nil), nil
-	}
-	errOutFor := func(project, resource string) (io.Writer, io.Closer, error) {
-		return stdErr, ioutil.NopCloser(nil), nil
-	}
-	task := ResourceDefinitions(project, []string{resource}, outFor, errOutFor)
+// getDumpedResourceAsStruct retrieves the requested resource from the dump directory and
+// decodes the JSON into the provided interface
+func getDumpedResource(path []string, dest interface{}) error {
+	filePath := filepath.Join(path...)
+	contents, err := ioutil.ReadFile(filePath)
 
-	err := task()
-	if err != nil {
-		return err
-	}
-
-	stdErrString := string(stdErr.Bytes())
-	if stdErrString != "" {
-		return errors.New(stdErrString)
-	}
-
-	decoder := json.NewDecoder(stdOut)
+	decoder := json.NewDecoder(bytes.NewBuffer(contents))
 	err = decoder.Decode(&dest)
 	if err != nil {
-		stdErr.Write([]byte(err.Error()))
 		return err
 	}
 
@@ -149,12 +121,13 @@ func getResourceStruct(project, resource string, dest interface{}) error {
 // are exhibiting signs that they have experienced an ImagePullBackOff recently
 // this will be reflected in the returned Result data. Any errors are written to
 // the supplied stdErr writer.
-func CheckImagePullBackOff(project string, stdErr io.Writer) (Result, error) {
+func CheckImagePullBackOff(project, basepath string) (Result, error) {
 	result := Result{Status: 0, StatusMessage: "this issue was not detected", CheckName: "check deploys for ImagePullBackOff error"}
 	events := Events{}
-	err := getResourceStruct(project, "events", &events)
+	err := getDumpedResource([]string{basepath, "projects", project, "definitions", "events.json"}, &events)
 	if err != nil {
-		stdErr.Write([]byte(err.Error()))
+		result.Status = 2
+		result.StatusMessage = "Error executing task"
 		return result, err
 	}
 
@@ -174,12 +147,13 @@ func CheckImagePullBackOff(project string, stdErr io.Writer) (Result, error) {
 // supplied project and if any have replicas set to zero this will be reflected
 // in the returned Result data. Any errors are written to the supplied stdErr
 // writer.
-func CheckDeployConfigsReplicasNotZero(project string, stdErr io.Writer) (Result, error) {
+func CheckDeployConfigsReplicasNotZero(project, basepath string) (Result, error) {
 	result := Result{Status: 0, StatusMessage: "this issue was not detected", CheckName: "check deployconfig replicas not 0"}
 	deploymentConfigs := DeploymentConfigs{}
-	err := getResourceStruct(project, "dc", &deploymentConfigs)
+	err := getDumpedResource([]string{basepath, "projects", project, "definitions", "deploymentconfigs.json"}, &deploymentConfigs)
 	if err != nil {
-		stdErr.Write([]byte(err.Error()))
+		result.Status = 2
+		result.StatusMessage = "Error executing task"
 		return result, err
 	}
 
--- a/analysis_test.go
+++ b/analysis_test.go
@@ -1,18 +1,9 @@
 package main
 
 import (
-	"bytes"
 	"errors"
-	"io"
-	"io/ioutil"
 	"testing"
 )
-
-var b *bytes.Buffer
-
-func mockOutFor(project, resource string) (io.Writer, io.Closer, error) {
-	return b, ioutil.NopCloser(nil), nil
-}
 
 func mockCheckFactoryOnePassOneFail() []CheckTask {
 	return []CheckTask{mockTestOne, mockTestTwo}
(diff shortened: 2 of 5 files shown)


analysis.go
```diff
@@ -50,60 +49,51 @@ type DeploymentConfigs struct {
...
 }
 
-type CheckTask func(string, io.Writer) (Result, error)
+type CheckTask func(string, string) (Result, error)
+
+func GetAnalysisTasks(tasks chan<- Task, basepath string, projects []string, results chan<- CheckResults) {
+	// Platform-wide analysis goes here
+
+	// project specific analysis in here
+	for _, p := range projects {
```
maleck13: not a big deal but wondering if you really need this function now?

analysis_test.go
```diff
@@ -1,62 +1,57 @@
...
+	task = checkProjectTask(mockCheckFactoryOnePass, "MockProject", "/", results)
 	err = task()
 
 	if err != nil {
 		t.Fatal("Expected no errors")
 	}
 
-	if len(string(b.Bytes())) == 0 {
-		t.Fatal("Tests not called")
+	close(results)
```
maleck13: this seems wrong. Closing the channel then reading from it. It should cause a panic

maleck13: same here

main.go
```diff
@@ -197,6 +198,8 @@ func main() {
 
 	runner := NewDumpRunner(basePath)
 
-	log.Print("Running tasks...")
-	RunAllTasks(runner, basePath, *concurrentTasks)
+	log.Print("Running dump tasks...")
+	RunAllDumpTasks(runner, basePath, *concurrentTasks)
```
maleck13: do we have to do the dereference here? As in does this have to be a pointer or could it just be a value? Not familiar with all the code so unsure. Again not a big deal

analysis_test.go
philbrookes: I thought that if I didn't do it this way around, the `for result := range results {` would block forever. This works as it has a buffer. Not sure how else I could test this without using goroutines?

tasks.go
```diff
@@ -137,19 +139,98 @@ func GetAllTasks(runner Runner, basepath string) <-chan Task {
...
+			for result := range checkResults {
+				analysisResults[result.Scope] = result.Results
+				output, err := json.MarshalIndent(analysisResults, "", "    ")
+				if err != nil {
+					results <- err
+				}
+				analysisFile.Write(output)
+			}
+		}()
 		wg.Wait()
```
maleck13: this seems a bit complex.  do we need the nested go func, not sure we do. Seems like it could even be a separate function?

analysis.go
philbrookes: There's not much here at the moment, but as more analysis tasks are added this function will grow.

tasks.go
```diff
@@ -137,19 +139,98 @@ func GetAllTasks(runner Runner, basepath string) <-chan Task {
...
-			errOutFor := outToFile(basepath, "stderr", "analysis")
-			tasks <- CheckTasks(p, outFor, errOutFor)
+	// Loop through the task execution results and log errors.
+	for err := range results {
+		taskCount++
+		if err != nil {
+			// TODO: there should be a way to identify which task
+			// had an error.
+			fmt.Fprintln(os.Stderr)
+			log.Printf("Task error: %v", err)
```
maleck13: does this log actually go to stderr ?

analysis_test.go
maleck13: actually I am wrong sending on a closed channel panics. This still seems a bit weird though. How do you know you have received anything on the channel when you close it? or is that ok

tasks.go
philbrookes: Yes

main.go
philbrookes: Yeah, this confused me too, it's the way the flags module parses the flags to the program.

analysis_test.go
philbrookes: As this is a unit test - if nothing is sent on the channel then the test should fail.

I can assert that I should have received something on this channel by the time I close it, as the code that sends into this channel runs synchronously, at line 28.

maleck13:
👍

david-martin:
@philbrookes as disucssed on irc, the output is appending to the analysis.json file at the moment with each result, rather than rewriting, resulting in a much larger file with repeat output

philbrookes:
Thanks @david-martin, I have addressed that issue now.

david-martin:
@philbrookes  there's something odd about the analysis.json file, different this time

it looks like a lot of 0000's (hex) are written at the start of the file before the json content starts. possible that the file rewrite isn't quiet rewriting but still appending in a weird way?

philbrookes:
@david-martin That was strange alright. I've changed the way it writes to the file now, I think how I was doing it earlier was not the correct method. Could you retry when you get a chance?

david-martin:
@philbrookes Tryed again and looks good.

```
cat rhmap-dumps/2016-08-23T13-32-46Z/analysis.json | jq
{
  "cloudappdevjuuv": [
    {
      "checkName": "check deploys for ImagePullBackOff error",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    },
    {
      "checkName": "check deployconfig replicas not 0",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    }
  ],
  "cloudapplivezfhr": [
    {
      "checkName": "check deploys for ImagePullBackOff error",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    },
    {
      "checkName": "check deployconfig replicas not 0",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    }
  ],
  "dmtestbuild01": [
    {
      "checkName": "check deploys for ImagePullBackOff error",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    },
    {
      "checkName": "check deployconfig replicas not 0",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    }
  ],
  "jsmolarlivequsq": [
    {
      "checkName": "check deploys for ImagePullBackOff error",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    },
    {
      "checkName": "check deployconfig replicas not 0",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    }
  ],
  "jsmolaruarttestsliveegbm": [
    {
      "checkName": "check deploys for ImagePullBackOff error",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    },
    {
      "checkName": "check deployconfig replicas not 0",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    }
  ],
  "logging": [
    {
      "checkName": "check deploys for ImagePullBackOff error",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    },
    {
      "checkName": "check deployconfig replicas not 0",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    }
  ],
  "m2-mbaas": [
    {
      "checkName": "check deploys for ImagePullBackOff error",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    },
    {
      "checkName": "check deployconfig replicas not 0",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    }
  ],
  "qe-3node-4-1": [
    {
      "checkName": "check deploys for ImagePullBackOff error",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    },
    {
      "checkName": "check deployconfig replicas not 0",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    }
  ],
  "qe-3node-rhmap410cr5-02": [
    {
      "checkName": "check deploys for ImagePullBackOff error",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    },
    {
      "checkName": "check deployconfig replicas not 0",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    }
  ],
  "rhmap-core": [
    {
      "checkName": "check deploys for ImagePullBackOff error",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    },
    {
      "checkName": "check deployconfig replicas not 0",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    }
  ],
  "testingcloudapplivetutp": [
    {
      "checkName": "check deploys for ImagePullBackOff error",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    },
    {
      "checkName": "check deployconfig replicas not 0",
      "status": 0,
      "statusMessage": "this issue was not detected",
      "info": null
    }
  ]
}
```

analysis.go
```diff
@@ -50,60 +49,51 @@ type DeploymentConfigs struct {
 	} `json:"items"`
 }
 
-type CheckTask func(string, io.Writer) (Result, error)
+type CheckTask func(string, string) (Result, error)
+
+func GetAnalysisTasks(tasks chan<- Task, basepath string, projects []string, results chan<- CheckResults) {
```
rhcarvalho: *Sorry for the late "review" on the merged code, just trying to get up-to-speed on what changed recently.*

Here, why is `results` an argument and not a return value?

analysis.go
```diff
@@ -50,60 +49,51 @@ type DeploymentConfigs struct {
...
 		return []CheckTask{CheckImagePullBackOff, CheckDeployConfigsReplicasNotZero}
-	}, project, outFor, errOutFor)
+	}, project, basepath, results)
 }
 
 // A getProjectCheckFactory generates tasks to diagnose system conditions.
 type getProjectCheckFactory func() []CheckTask
 
 type CheckResults struct {
+	Scope   string `json:"scope"`
```
rhcarvalho: What is the `Scope` for? I'm missing some docs.
Why do we make it `"scope"` in JSON, but let `Results` use the default value (capitalized name)?

rhcarvalho: One needs to dig the code to see that Scope == project :crying_cat_face:

analysis.go
```diff
@@ -50,60 +49,51 @@ type DeploymentConfigs struct {
...
-		}
-		defer stdOutCloser.Close()
-		stdErr, stdErrCloser, err := errOutFor(project, "analysis")
-		if err != nil {
-			return err
-		}
-		defer stdErrCloser.Close()
-
-		results := CheckResults{Results: []Result{}}
-
```
rhcarvalho: Doesn't it feel good to get rid of code? :)

:+1: to the refactor in general.

analysis.go
```diff
@@ -112,33 +102,15 @@ func checkTasks(checkFactory getProjectCheckFactory, project string, outFor, err
...
-	err := task()
-	if err != nil {
-		return err
-	}
-
-	stdErrString := string(stdErr.Bytes())
-	if stdErrString != "" {
-		return errors.New(stdErrString)
-	}
+// getDumpedResourceAsStruct retrieves the requested resource from the dump directory and
```
rhcarvalho: doc comment doesn't match the method name.

analysis.go
```diff
@@ -112,33 +102,15 @@ func checkTasks(checkFactory getProjectCheckFactory, project string, outFor, err
...
+// decodes the JSON into the provided interface
+func getDumpedResource(path []string, dest interface{}) error {
+	filePath := filepath.Join(path...)
+	contents, err := ioutil.ReadFile(filePath)
 
-	decoder := json.NewDecoder(stdOut)
+	decoder := json.NewDecoder(bytes.NewBuffer(contents))
```
rhcarvalho: This is begging to be `json.Unmarshal` instead; see https://golang.org/pkg/encoding/json/#Unmarshal.

analysis.go
```diff
@@ -112,33 +102,15 @@ func checkTasks(checkFactory getProjectCheckFactory, project string, outFor, err
...
 	err = decoder.Decode(&dest)
 	if err != nil {
-		stdErr.Write([]byte(err.Error()))
```
rhcarvalho: You can replace the whole surrounding area here with:

```go
return json.Unmarshal(contents, &dest)
```

analysis.go
```diff
@@ -112,33 +102,15 @@ func checkTasks(checkFactory getProjectCheckFactory, project string, outFor, err
...
-	}
-
-	stdErrString := string(stdErr.Bytes())
-	if stdErrString != "" {
-		return errors.New(stdErrString)
-	}
+// getDumpedResourceAsStruct retrieves the requested resource from the dump directory and
+// decodes the JSON into the provided interface
+func getDumpedResource(path []string, dest interface{}) error {
+	filePath := filepath.Join(path...)
```
rhcarvalho: I'd suggest just taking a path, string, since that's what we want and need here. The caller constructs the path in whatever way it wants.

In the end of the day, this function is just a strange wrapper around what `json.Unmarshal` does. But the usage of `ioutil.ReadFile` is also suspicious... the normal thing to do is to open the file with `os.Open`, pass it to `json.NewDecoder`, then decode.

analysis.go
```diff
@@ -174,12 +147,13 @@ func CheckImagePullBackOff(project string, stdErr io.Writer) (Result, error) {
...
 // writer.
-func CheckDeployConfigsReplicasNotZero(project string, stdErr io.Writer) (Result, error) {
+func CheckDeployConfigsReplicasNotZero(project, basepath string) (Result, error) {
 	result := Result{Status: 0, StatusMessage: "this issue was not detected", CheckName: "check deployconfig replicas not 0"}
 	deploymentConfigs := DeploymentConfigs{}
-	err := getResourceStruct(project, "dc", &deploymentConfigs)
+	err := getDumpedResource([]string{basepath, "projects", project, "definitions", "deploymentconfigs.json"}, &deploymentConfigs)
 	if err != nil {
-		stdErr.Write([]byte(err.Error()))
+		result.Status = 2
```
rhcarvalho: Where does this `2` comes from?!

analysis.go
```diff
@@ -174,12 +147,13 @@ func CheckImagePullBackOff(project string, stdErr io.Writer) (Result, error) {
...
+		result.StatusMessage = "Error executing task"
```
rhcarvalho: Why not include the error itself in the result struct?

analysis_test.go
```diff
@@ -1,62 +1,57 @@
...
 
 	if err == nil {
 		t.Fatal("Expected error")
 	}
 
-	if len(string(b.Bytes())) == 0 {
-		t.Fatal("Tests not called")
-
+	result := <-results
+	if (len(result.Results)) != 2 {
```
rhcarvalho: Naming is strange here... a "result" has multiple "results" within?!

dump_writers.go
```diff
@@ -27,3 +27,7 @@ func outToFile(basepath, extension, scope string) projectResourceWriterCloserFac
 		return f, f, nil
 	}
 }
+
+// A pathResourceWriterFactory generates io.Writers for dumping data of a
+// particular resource type within a project.
+type pathResourceWriterCloserFactory func() (io.Writer, io.Closer, error)
```
rhcarvalho: This is not used anywhere...

main.go
```diff
@@ -188,6 +188,7 @@ func main() {
 		// error from os.RemoveAll is intentionally ignored, since there
 		// is no useful action we can do, and we don't need to confuse
 		// the user with an error message.
+
```
rhcarvalho: This blank line is separating code from a comment that explains it. I think it doesn't belong here.

main.go
rhcarvalho: `flag.Int` & family return pointer values.
"If you like, you can bind the flag to a variable using the Var() functions." ([pkg/flag docs](https://golang.org/pkg/flag/))

Because `concurrentTasks` is a `*int`, we need to dereference because we want an `int` type in the argument.

`flag.IntVar` takes a pointer because it needs to mutate the value. And `flag.Int` returns a pointer for similar reasons, otherwise you'd need to have the flag value determined at the time you declare the flag, which doesn't make sense.

main.go
```diff
@@ -197,6 +198,8 @@ func main() {
 
 	runner := NewDumpRunner(basePath)
 
-	log.Print("Running tasks...")
-	RunAllTasks(runner, basePath, *concurrentTasks)
+	log.Print("Running dump tasks...")
```
rhcarvalho: Do we really want to bother users about "dump" x "analysis" tasks? Those seem like internal implementation details to me.

The idea of printing "Doing XYZ" and then a "." every now and then was just as a poor man's progress bar.

tasks.go
```diff
@@ -137,19 +140,96 @@ func GetAllTasks(runner Runner, basepath string) <-chan Task {
...
+	}()
+	return tasks
+}
 
+func RunAllAnalysisTasks(runner Runner, path string, workers int) {
+	start := time.Now()
+
+	checkResults := make(chan CheckResults)
+	tasks := GetAllAnalysisTasks(runner, path, checkResults)
+	results := make(chan error)
```
rhcarvalho: Why do we need to convert from a `chan CheckResults` to a `chan error`?

tasks.go
```diff
@@ -137,19 +140,96 @@ func GetAllTasks(runner Runner, basepath string) <-chan Task {
...
+			defer wg.Done()
+			for task := range tasks {
+				results <- task()
+			}
+		}()
+	}
+
+	// Listen to the checkResults channel and write all the results into
+	// the analysis.json file
+	go func() {
```
rhcarvalho: I wonder if something goes wrong here, aren't we risking truncating the file?
IIUC we're overwriting the same file over and over.

The idea seems nice, but perhaps it is safer to write to a different file and move it later? I think `vim` and some other editors and utilities take that approach?

analysis.go
philbrookes: good question, I just picked a non-zero number to indicate a problem... this could probably do with being defined somewhere though, and maybe more useful codes that could be searched in documentation or something?

analysis.go
philbrookes: The check tasks run asynchronously, so they send their results down the results channel that is passed in here.

analysis.go
philbrookes: This code felt much better by the time this PR was finished. Can't wait to improve it even more!

analysis_test.go
philbrookes: Sure, I'm open to other suggestions here, the idea is like so:
- A task runs against a project (or against the platform, though we have none of these implemented yet).
- The task runs several checks inside that project synchronously.
- Each check has a result
- Each task has a result which contains the check results.

Would it make more sense to have something like:
```
taskResult := <-taskResults
if(len(taskResult.CheckResults)) != 2 {
    ...
```

tasks.go
philbrookes: not sure I understand the suggestion here. Do you mean that each iteration creates a new file? and if the loop exits successfully, take the last file (or maybe the largest file) and make that into analysis.json and delete all the older ones?

analysis.go
rhcarvalho: I understand we need a results channel, but why does the caller need to create the channel and pass it in as an argument?

Granted I might be missing something, I thought of `GetAnalysisTasks` contract as "I'll send tasks to a channel, and here's where you can look for the results of those tasks if you want".

Generally speaking, I felt there is some difficulty to arrange for closing the results channel... how to tell that no more results will be sent. I might need to reason more about https://github.com/feedhenry/fh-system-dump-tool/pull/36/files/9ef911e9525d4008c952979fbf3bfd5514455aa0#diff-2628f4a06e8be3343e85abecc49fae35R193

analysis_test.go
rhcarvalho: > The task runs several checks inside that project synchronously.

Does it need to be synchronous?

---

It might be worth exploring that idea. In my head I was trying to get to task == check, so that we have a simple execution model: concurrently put tasks to a "bag" and execute them.

For things that we want to execute per project, there would be as many of those tasks as projects (so that they can be executed concurrently).

I see the original Task signature, `func() error` (is it?), doesn't help us grouping results as we want in the "analysis tasks"....

tasks.go
rhcarvalho: I meant something along the lines of

```go
// note that this is a bad choice of variable name, as it shadows the package filepath
filepath := filepath.Join(path, "analysis.json.tmp")
// ...
os.Rename(...)
```

rhcarvalho: Only rename after a successful write. Notice we're ignoring the error from `ioutil.WriteFile`.

PR #37, opened 2016-08-23
check event log for errors, and improve tests on analysis

philbrookes:
# Motivation
The dump tool is intended to check the event log for any errors and add these to the analysis results.

The analysis in the dump tool is not well tested currently.

# Changes
Alter an existing analysis check to check for any errors in the event log, rather than specific errors

Update the analysis code to be more testable.

Add unit tests to the test suite to test the analysis checks.

# Jiras
[RHMAP-9453](https://issues.jboss.org/browse/RHMAP-9453)

### DIFF
--- a/analysis.go
+++ b/analysis.go
@@ -1,11 +1,9 @@
 package main
 
 import (
-	"bytes"
 	"encoding/json"
-	"io/ioutil"
+	"os"
 	"path/filepath"
-	"strings"
 )
 
 type Info struct {
@@ -17,23 +15,27 @@ type Info struct {
 }
 
 type Result struct {
-	CheckName     string `json:"checkName" yaml:"checkName"`
-	Status        int    `json:"status" yaml:"status"`
-	StatusMessage string `json:"statusMessage" yaml:"statusMessage"`
-	Info          []Info `json:"info" yaml:"info"`
+	CheckName     string  `json:"checkName" yaml:"checkName"`
+	Status        int     `json:"status" yaml:"status"`
+	StatusMessage string  `json:"statusMessage" yaml:"statusMessage"`
+	Info          []Info  `json:"info" yaml:"info"`
+	Events        []Event `json:"events" yaml:"events"`
+}
+
+type Event struct {
+	Kind           string `json:"kind"`
+	InvolvedObject struct {
+		Namespace string `json:"namespace"`
+		Name      string `json:"name"`
+	} `json:"involvedObject"`
+	Reason  string `json:"reason"`
+	Message string `json:"message"`
+	Count   int    `json:"count"`
+	Type    string `json:"type"`
 }
 
 type Events struct {
-	Items []struct {
-		Kind           string `json:"kind"`
-		InvolvedObject struct {
-			Namespace string `json:"namespace"`
-			Name      string `json:"name"`
-		} `json:"involvedObject"`
-		Reason  string `json:"reason"`
-		Message string `json:"message"`
-		Count   int    `json:"count"`
-	} `json:"items"`
+	Items []Event `json:"items"`
 }
 
 type DeploymentConfigs struct {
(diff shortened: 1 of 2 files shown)


analysis.go
```diff
@@ -102,55 +106,57 @@ func checkProjectTask(checkFactory getProjectCheckFactory, project, basepath str
...
+// which will be parsed and loaded into the supplied interface
+type DumpedJSONResourceFactory func([]string, interface{}) error
+
+// Returns a factory which will load resource from a given basepath. The factory parses the file
+// contents as JSON and loads it into the provided dest interface
+func getDumpedJSONResourceFactory(basepath []string) DumpedJSONResourceFactory {
+	return func(path []string, dest interface{}) error {
+		file := filepath.Join(append(basepath, path...)...)
+		contents, err := ioutil.ReadFile(file)
+
```
maleck13: missing an err check here

analysis.go
```diff
@@ -102,55 +106,57 @@ func checkProjectTask(checkFactory getProjectCheckFactory, project, basepath str
...
+		decoder := json.NewDecoder(bytes.NewBuffer(contents))
+		err = decoder.Decode(&dest)
```
maleck13: bit more idiomatic to do 
```
if err := decoder.Decode(); err != nil{
return err
}

File: analysis.go [hunk 3]
```diff
@@ -102,55 +106,57 @@ func checkProjectTask(checkFactory getProjectCheckFactory, project, basepath str
...
-// the supplied stdErr writer.
-func CheckImagePullBackOff(project, basepath string) (Result, error) {
-	result := Result{Status: 0, StatusMessage: "this issue was not detected", CheckName: "check deploys for ImagePullBackOff error"}
+// CheckEventLogForErrors checks all events in the supplied project and if any
+// are not type 'Normal' (i.e. Warning or Error), it will add them to the returned results.
+func CheckEventLogForErrors(JSONResourceFactory DumpedJSONResourceFactory) (Result, error) {
+	result := Result{Status: 0, StatusMessage: "this issue was not detected", CheckName: "check eventlog for any errors"}
 	events := Events{}
-	err := getDumpedResource([]string{basepath, "projects", project, "definitions", "events.json"}, &events)
+	err := JSONResourceFactory([]string{"definitions", "events.json"}, &events)
```
maleck13: could also be  changed to  ``` if err:= ```

analysis.go
```diff
@@ -102,55 +106,57 @@ func checkProjectTask(checkFactory getProjectCheckFactory, project, basepath str
...
 	if err != nil {
 		result.Status = 2
 		result.StatusMessage = "Error executing task"
 		return result, err
 	}
 
 	for _, event := range events.Items {
-		if event.Reason == "FailedSync" && strings.Contains(event.Message, "ImagePullBackOff") {
-			info := Info{Name: event.InvolvedObject.Name, Namespace: event.InvolvedObject.Namespace, Kind: event.Kind, Count: event.Count, Message: event.Message}
+		if event.Type != "Normal" {
```
maleck13: What are the possible types? assuming they are different than the Reasons

analysis_test.go
```diff
@@ -55,3 +61,147 @@ func TestCheckTasksWithPass(t *testing.T) {
...
+	if res.Status != 1 {
+		t.Fatal("res.Status expected: 1, got:", res.Status)
+	}
+	if len(res.Events) != 1 {
+		t.Fatal("len(res.Events) expected: 1, got:", len(res.Events))
+	}
+	if res.Events[0].Count != 94215 {
+		t.Fatal("res.Events[0].Count expected: 94215, got:", string(res.Events[0].Count))
+	}
+	if res.Events[0].Type != "Warning" {
```
maleck13: hmm not sure what we are testing here. I guess that we can decode the json into a the correct struct ?

analysis.go
maleck13: Any reason to read the whole thing into mem? Would opening the file and passing the file object that is a reader to  the decoder be an option. Not sure what the encoder does underneath there. Likely has to read the whole thing cause json can't be partially parsed. So there may not be a benefit

maleck13:
reviewed

analysis.go
philbrookes: The Types are the likes of: 'Normal', 'warning', etc... Reasons are more like 'FailedSync' or 'DeploymentCreated', etc.

analysis_test.go
philbrookes: yeah, and that the analysis check preserves the data from the event log correctly. I'm not sure how useful these are either, but figured I'd pop them in anyway.

analysis.go
philbrookes: cool, good point, I'll update this now.

maleck13:
👍

david-martin:
@philbrookes Verified that an `events` is included for each project https://gist.github.com/david-martin/a75b78b1edbfd4bdd5cb5195f51ca33f
Minor thing, would it make sense to set `events` to an empty array if there are no events rather than `null`?
I think `null` makes sense if its meant to be a string, number or object, but doesn't have a value.

philbrookes:
@david-martin good idea, I've done that now, and also for the info array as well. I'll merge once the checks pass.

PR #39, opened 2016-08-26
Fixed specification of project

jhellar:
# Motivation
Currently project is not specified for commands retrieving data from nagios and millicore. Because of that they will only work if there is only one project with these components and it is the oc's current project. Otherwise project have to be specified in the commands.

# Changes
Added specification for project.


jhellar:
Not yet finished. Tests need to be fixed as well.

jhellar:
Tests fixed.

philbrookes:
:+1: Looks good

PR #40, opened 2016-08-26
Rephrase output when archiving the dumped data fails

philbrookes:
# Motivation
Make it clear to the user that although the archive was not created, the dump data is still accessible, and where it is.

# Changes
Rephrased output when the tar command fails.

### DIFF
--- a/main.go
+++ b/main.go
@@ -180,7 +180,7 @@ func main() {
 		cmd.Stdout = &stdout
 		cmd.Stderr = &stderr
 		if err := cmd.Run(); err != nil {
-			log.Printf("Dumped system information to: %s", basePath)
+			log.Printf("There was an error creating archiving the dumped data, so the dumped system information is stored unarchived in: %s", basePath)
 			return
 		}
 


jhellar:
LGTM

main.go
```diff
@@ -180,7 +180,7 @@ func main() {
 		cmd.Stdout = &stdout
 		cmd.Stderr = &stderr
 		if err := cmd.Run(); err != nil {
-			log.Printf("Dumped system information to: %s", basePath)
+			log.Printf("There was an error creating archiving the dumped data, so the dumped system information is stored unarchived in: %s", basePath)
```
rhcarvalho: "creating archiving" looks like a typo?

rhcarvalho:
I don't understand much the motivation. Why do we need to talk about "failed to create archive", when we are in fact failing gracefully?

There is no action the user can take when the tar step fails.

jhellar:
I think that the user should be notified that creating of archive failed because if not user can be confused when support team will ask him for archive.

rhcarvalho:
Well, how does the message help users in that case? The message as is doesn't help users create an archive on their own.

Perhaps this could be a small note in the docs of the tool. "If you don't have a TIMESTAMP.tar.gz file, but instead a TIMESTAMP/ directory, then it means that ...; please when requested to send data to support, archive the directory using a format such as zip, tar.gz, etc".

If we can create files etc, there's not much reason for "tar" to fail, except for `tar` must be available in the PATH... for that the solution is changing to a pure-Go solution.

PR #41, opened 2016-08-31
Continue with dump even when there is no project

jhellar:
# Motivation

Currently when there is no project the dump tool will not run any dump task. There are dump tasks which are platform-wide so these should be run even when there is no project.
# Change

Removed return line in no projects check.

### DIFF
--- a/tasks.go
+++ b/tasks.go
@@ -79,7 +79,6 @@ func GetAllDumpTasks(runner Runner, basepath string) <-chan Task {
 		}
 		if len(projects) == 0 {
 			tasks <- NewError(errors.New("no projects visible to the currently logged in user"))
-			return
 		}
 
 		var wg sync.WaitGroup


rhcarvalho:
The `return` should apply only to adding tasks that depend on having a list of projects.
We don't need to add them when the list is empty.

The only thing that seems to be out of that is the `GetOcAdmDiagnosticsTask`, I'd move that out to another place, since it doesn't depend on the list of projects.

rhcarvalho:
Hmm, perhaps removing the `return` is the way to go, I see `GetOpenShiftMetadataTasks` currently has two parts, one that uses projects, one that doesn't.

So unless we do a bigger restructuring, removing the `return` seems fine.

rhcarvalho:
In any case, @jhellar look at [line 78](https://github.com/feedhenry/fh-system-dump-tool/pull/41/files#diff-2628f4a06e8be3343e85abecc49fae35R78), there we're also returning when `oc get projects` fails... so maybe we get rid of both `return` statements?

rhcarvalho:
@jhellar ping

jhellar:
@rhcarvalho sorry for the late response, I'm not sure about the `return` statement at line 78. In that case it will return if there is an error using oc command. IMO if it won't return probably all tasks will fail. WDYT?

rhcarvalho:
In both cases we could equally proceed to run tasks that do not depend on the list of projects.

But given the code freeze, unless that's a bug, we should revisit it later?

jhellar:
OK, makes sense.

PR #45, opened 2016-09-07
Take version from Git in build time

rhcarvalho:
As part of having a standard build environment for building releases, this PR introduces a Makefile to use for "standard builds".

The code base is still buildable with the standard go tool without any special flag, though using `make` we consistently add version information taken from Git and loaded into the built binary.

Note that this doesn't yet give us a build environment for community releases, this is something we may investigate separately (e.g., using Travis and integrating with GitHub).

--- 

https://issues.jboss.org/browse/RHMAP-9400

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,12 @@
+ifndef FH_SYSTEM_DUMP_TOOL_VERSION
+FH_SYSTEM_DUMP_TOOL_VERSION := $(shell git describe --tags --abbrev=14)
+endif
+LDFLAGS := -X main.Version=$(FH_SYSTEM_DUMP_TOOL_VERSION)
+
+.PHONY: all
+all:
+	@go install -v -ldflags '$(LDFLAGS)'
+
+.PHONY: clean
+clean:
+	@-go clean -i
--- a/main.go
+++ b/main.go
@@ -155,7 +155,7 @@ func main() {
 
 	var b bytes.Buffer
 	PrintVersion(&b)
-	if err := ioutil.WriteFile(filepath.Join(basePath, ".version"), b.Bytes(), 0660); err != nil {
+	if err := ioutil.WriteFile(filepath.Join(basePath, "version"), b.Bytes(), 0660); err != nil {
 		log.Fatalln("Error:", err)
 	}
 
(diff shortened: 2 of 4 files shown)


main.go
```diff
@@ -156,7 +153,9 @@ func main() {
 		log.Fatalln("Error:", err)
 	}
 
-	if err := ioutil.WriteFile(filepath.Join(basePath, ".version"), []byte(version), 0660); err != nil {
+	var b bytes.Buffer
+	PrintVersion(&b)
+	if err := ioutil.WriteFile(filepath.Join(basePath, "version"), b.Bytes(), 0660); err != nil {
```
philbrookes: Do you think this would be better in the metadata file?

philbrookes:
:+1: Verified this locally and it looks fine.

main.go
rhcarvalho: It could be, but I think one doesn't exist yet?

Makefile
```diff
@@ -0,0 +1,12 @@
+ifndef FH_SYSTEM_DUMP_TOOL_VERSION
```
rhcarvalho: This allows us to override the version during build, either via environment variable or extra argument to `make`.

```bash
# 1
make FH_SYSTEM_DUMP_TOOL_VERSION=v1.0.0

# 2
FH_SYSTEM_DUMP_TOOL_VERSION=v1.0.0 make

# 3
export FH_SYSTEM_DUMP_TOOL_VERSION=v1.0.0
make
```

@philbrookes @vibe13, I went with a rather long variable name to namespace it properly (instead of calling it `VERSION`) without introducing alternative names or acronyms to the tool (thus not calling it `FH_SDT_VERSION`, for instance). I'm open to naming suggestions if there is a pattern to be followed.

main.go
philbrookes: oh yes, I was thinking of: https://github.com/feedhenry/fh-system-dump-tool/blob/master/openshift_metadata.go

but I see now that is only for openshift metadata, not general metadata.

Makefile
philbrookes: This is a great idea.

main.go
rhcarvalho: > Do you think this would be better in the metadata file?

Hmm, did you mean Go source code file or dump output file?

philbrookes: I meant dump output file.

PR #46, opened 2016-09-07
output from analysis - RHMAP-9595

philbrookes:
# Motivation
When dump and analyse is complete, output any discovered issues to the user on stdout. 

# Changes
Add task to run after analysis.json is finished being written to output any discovered errors.

Example output:
```
Potential issue discovered in rhmap-core: check eventlog for any errors
  Details:
    Readiness probe failed: Get http://10.1.6.13:8080/sys/info/ping: dial tcp 10.1.6.13:8080: connection refused
    Cannot update deployment rhmap-core/fh-supercore-3 status to Pending: replicationcontrollers "fh-supercore-3" cannot be updated: the object has been modified; please apply your changes to the latest version and try again
Potential issue discovered in testingcloudapp1oldrr64: check eventlog for any errors
  Details:
    Cannot update deployment testingcloudapp1oldrr64/nodejs-frontend-1 status to Pending: replicationcontrollers "nodejs-frontend-1" cannot be updated: the object has been modified; please apply your changes to the latest version and try again
    Cannot update deployment testingcloudapp1oldrr64/redis-1 status to Pending: replicationcontrollers "redis-1" cannot be updated: the object has been modified; please apply your changes to the latest version and try again
Potential issue discovered in testingcloudappolderr64: check eventlog for any errors
  Details:
    Cannot update deployment testingcloudappolderr64/nodejs-frontend-1 status to Pending: replicationcontrollers "nodejs-frontend-1" cannot be updated: the object has been modified; please apply your changes to the latest version and try again
    Cannot update deployment testingcloudappolderr64/redis-1 status to Pending: replicationcontrollers "redis-1" cannot be updated: the object has been modified; please apply your changes to the latest version and try again
2016/09/07 15:54:25 Dumped system information to: rhmap-dumps/2016-09-07T14-53-57Z.tar.gz
```

This has discovered errors in 3 different projects:
-  rhmap-core
- testingcloudapp1oldrr64
- testingcloudappolderr64

I am not sure if this format is the most ideal, and I am open to suggestions, the problem really seems to be that there is quite a lot of text to output. So possibly it could look more brief like:
```
Potential issue discovered in rhmap-core: check eventlog for any errors
   For more information check the analysis.json in the archive
Potential issue discovered in testingcloudapp1oldrr64: check eventlog for any errors
   For more information check the analysis.json in the archive
Potential issue discovered in testingcloudappolderr64: check eventlog for any errors
   For more information check the analysis.json in the archive
```

Which is less immediately useful, but is far less clutter on the screen. Any thoughts?

### DIFF
--- a/analysis.go
+++ b/analysis.go
@@ -2,8 +2,14 @@ package main
 
 import (
 	"encoding/json"
-	"os"
+	"io/ioutil"
 	"path/filepath"
+)
+
+const (
+	analysisErrorNotDiscovered = iota
+	analysisErrorReadingDumpedResource
+	analysisErrorDiscoveredByAnalysis
 )
 
 // Info is a piece of information regarding a check, multiple Info can be
@@ -95,6 +101,9 @@ type Pods struct {
 // CheckTask is the interface which checks must implement.
 type CheckTask func(DumpedJSONResourceFactory) (Result, error)
 
+// AnalysisResults is a representation of the JSON analysis results file.
+type AnalysisResults map[string]map[string][]Result
+
 // GetAnalysisTasks creates all the analysis tasks and sends them one by one
 // down the tasks Channel.
 func GetAnalysisTasks(tasks chan<- Task, basepath string, projects []string, results chan<- CheckResults) {
@@ -102,7 +111,7 @@ func GetAnalysisTasks(tasks chan<- Task, basepath string, projects []string, res
 
 	// project specific analysis in here.
 	for _, p := range projects {
-		JSONResourceFactory := getDumpedJSONResourceFactory([]string{basepath, "projects", p})
+		JSONResourceFactory := getDumpedJSONResourceFactory(filepath.Join(basepath, "projects", p))
 		tasks <- CheckProjectTask(p, results, JSONResourceFactory)
 	}
 }
@@ -121,6 +130,7 @@ type getProjectCheckFactory func() []CheckTask
 // project or platform-wide.
 type CheckResults struct {
 	Scope   string `json:"scope"`
+	Name    string `json:"name,omitempty"`
 	Results []Result
 }
 
@@ -129,7 +139,7 @@ type CheckResults struct {
 // combined into a single JSON object and returned.
 func checkProjectTask(checkFactory getProjectCheckFactory, JSONResourceFactory DumpedJSONResourceFactory, project string, results chan<- CheckResults) Task {
 	return func() error {
-		result := CheckResults{Scope: project, Results: []Result{}}
+		result := CheckResults{Scope: "project", Name: project, Results: []Result{}}
 		checks := checkFactory()
 
 		var errors errorList
@@ -151,34 +161,29 @@ func checkProjectTask(checkFactory getProjectCheckFactory, JSONResourceFactory D
 	}
 }
 
-// DumpedJSONResourceFactory takes an array of strings describing the path to a
+// DumpedJSONResourceFactory takes a string containing the path to a
 // JSON file which will be parsed and loaded into the supplied interface.
-type DumpedJSONResourceFactory func([]string, interface{}) error
+type DumpedJSONResourceFactory func(string, interface{}) error
 
 // getDumpedJSONResourceFactory returns a factory which will load resource from
 // a given basepath. The factory parses the file contents as JSON and loads it
 // into the provided dest interface.
-func getDumpedJSONResourceFactory(basepath []string) DumpedJSONResourceFactory {
-	return func(path []string, dest interface{}) error {
-		file := filepath.Join(append(basepath, path...)...)
-		contents, err := os.Open(file)
+func getDumpedJSONResourceFactory(basepath string) DumpedJSONResourceFactory {
+	return func(path string, dest interface{}) error {
+		contents, err := ioutil.ReadFile(filepath.Join(basepath, path))
 		if err != nil {
 			return err
 		}
-		decoder := json.NewDecoder(contents)
-		if err := decoder.Decode(&dest); err != nil {
-			return err
-		}
-		return nil
+		return json.Unmarshal(contents, dest)
 	}
 }
 
 // CheckForWaitingPods checks all pods for any containers in waiting status.
 func CheckForWaitingPods(JSONResourceFactory DumpedJSONResourceFactory) (Result, error) {
-	result := Result{Status: 0, StatusMessage: "this issue was not detected", CheckName: "check pods for 'waiting' containers", Info: []Info{}, Events: []Event{}}
+	result := Result{Status: analysisErrorNotDiscovered, StatusMessage: "this issue was not detected", CheckName: "check pods for 'waiting' containers", Info: []Info{}, Events: []Event{}}
 	pods := Pods{}
-	if err := JSONResourceFactory([]string{"definitions", "pods.json"}, &pods); err != nil {
-		result.Status = 2
+	if err := JSONResourceFactory(filepath.Join("definitions", "pods.json"), &pods); err != nil {
+		result.Status = analysisErrorReadingDumpedResource
 		result.StatusMessage = "Error executing task: " + err.Error()
 		return result, err
 	}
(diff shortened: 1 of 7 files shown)


philbrookes:
@david-martin @pmccarthy @mikenairn @jessesarnovsky @rhcarvalho Any thoughts on the output formatting discussed above?

rhcarvalho:
The verbosity of the output depends on how we expect people to consume the information... Are we aiming that people will open the archive to look for things there themselves, or is the archive something that only support cares about?


We could also experiment with grouping by "problem", if that makes sense?

```
Pods with errors in the event log:
  Project: foo
  Pods: x, y, z, ...

Another error:
  lorem ipsum...
```


And we need to think about errors that are not part of any project, is that covered?

analysis.go
```diff
@@ -2,10 +2,16 @@ package main
 
 import (
 	"encoding/json"
-	"os"
+	"io/ioutil"
 	"path/filepath"
 )
 
+var (
```
rhcarvalho: const?

analysis.go
```diff
@@ -2,10 +2,16 @@ package main
...
+	analysisErrorNotDiscovered         = 0
+	analysisErrorReadingDumpedResource = 1
```
rhcarvalho: iota? https://golang.org/ref/spec#Iota

analysis.go
```diff
@@ -2,10 +2,16 @@ package main
 
 import (
 	"encoding/json"
-	"os"
+	"io/ioutil"
 	"path/filepath"
 )
 
+var (
+	analysisErrorNotDiscovered         = 0
```
rhcarvalho: Does this refer to an error or non-error condition?

main.go
```diff
@@ -197,8 +197,15 @@ func main() {
...
-	log.Print("Running analysis tasks...")
 	RunAllAnalysisTasks(runner, basePath, *concurrentTasks)
+
+	delta := time.Since(start)
+	// Remove sub-second precision.
+	delta -= delta % time.Second
+	log.Printf("Run all tasks in %v.", delta)
+
+	RunOutputTask(runner, basePath, *concurrentTasks)
+
```
rhcarvalho: extraneous new line

philbrookes:
@rhcarvalho I had intended for errors that are not part of a project (there are currently no implemented analysis of this type) would have `scope` of `platform`

analysis.go
philbrookes: I considered using a const, but doesn't that make code that uses this value harder to test?

analysis.go
philbrookes: Yeah, I read about iota yesterday, but I thought they only worked with const's!

rhcarvalho:
> @rhcarvalho I had intended for errors that are not part of a project (there are currently no implemented analysis of this type) would have scope of platform

Does that mean we'll get confused if a project is named "platform"? :smile:

analysis.go
rhcarvalho: Why harder? I don't get it.

analysis.go
rhcarvalho: How about we call it some positive name, instead of double negative?

analysisOk?

main.go
```diff
@@ -197,8 +197,15 @@ func main() {
 
 	runner := NewDumpRunner(basePath)
 
-	log.Print("Running dump tasks...")
+	log.Print("Running dump and analyse tasks...")
```
rhcarvalho: English question.. `analyse` or `analysis` or sth else? :)

output_results.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import (
+	"strings"
+	"io"
+)
+
+// RunOutputTask will read in the analysis.json file and output
+// any useful information to o or any errors to e.
+func RunOutputTask(o, e io.Writer, runner Runner, resourceFactory DumpedJSONResourceFactory, workers int) {
```
rhcarvalho: Why do we write it to 2 different writers?

Seems that so far `e` is never used?

output_results.go
```diff
@@ -0,0 +1,28 @@
...
+	analysisResults := AnalysisResults{}
+	resourceFactory("analysis.json", &analysisResults)
+
+	for scope, results := range analysisResults {
+		for _, result := range results {
+			if result.Status != 0 {
+				o.Write([]byte("Potential issue discovered in " + scope + ": " + result.CheckName + "\n"))
```
rhcarvalho: This works, but perhaps would be cleaner to use `fmt.Fprintf`?

```go
fmt.Fprintf(o, "Potential issue discovered in %s: %s\n", scope, result.CheckName)
```

output_results.go
```diff
@@ -0,0 +1,28 @@
...
+
+// RunOutputTask will read in the analysis.json file and output
+// any useful information to o or any errors to e.
+func RunOutputTask(o, e io.Writer, runner Runner, resourceFactory DumpedJSONResourceFactory, workers int) {
+	analysisResults := AnalysisResults{}
+	resourceFactory("analysis.json", &analysisResults)
+
+	for scope, results := range analysisResults {
+		for _, result := range results {
+			if result.Status != 0 {
```
rhcarvalho: Please replace 0 with the constant it stands for.

tasks.go
```diff
@@ -167,8 +154,11 @@ func RunAllAnalysisTasks(runner Runner, path string, workers int) {
 
 	// Listen to the checkResults channel and write all the results into
 	// the analysis.json file
+	var writeWait sync.WaitGroup
```
rhcarvalho: Maybe this function is doing too much to have 2 WaitGroups...

output_results.go
philbrookes: yeah this is much tidier! I'll update it now

rhcarvalho:
@philbrookes wouldn't it be easier if we didn't have to load JSON from a file?
One way or another, we already have the results as a Go struct/slice/idk, then we serialize it, and deserialize it back...

output_results.go
```diff
@@ -11,19 +12,19 @@ func RunOutputTask(o io.Writer, e io.Writer, resourceFactory DumpedJSONResourceF
...
 
 	for scope, results := range analysisResults {
 		for _, result := range results {
 			if result.Status != 0 {
-				o.Write([]byte("Potential issue discovered in " + scope + ": " + result.CheckName + "\n"))
+				fmt.Fprintf(o, "Potential issue discovered in %s: %s\n", scope, result.CheckName)
 				o.Write([]byte("  Details:\n"))
 				for _, info := range result.Info {
-					o.Write([]byte("    " + strings.Replace(strings.TrimSpace(info.Message), "\n", "\n    ", -1) + "\n"))
+					fmt.Fprintf(o, "    %s\n", strings.Replace(strings.TrimSpace(info.Message), "\n", "\n    ", -1))
```
rhcarvalho: What are we trying to accomplish here with strings.TrimSpace and strings.Replace?
I wonder if we can get the same outcome without so much manipulation?

main.go
philbrookes: hmm, I'm not sure... I think you analyse something to get an analysis as the result? So analyse is the action and analysis is the object? 

I'm getting semantic satiation on this word now!

output_results.go
philbrookes: The messages that come out of the event log have newlines in them. So this is replace the newlines with indented newlines that match the formatted output.

philbrookes: The trim is because some of the messages in the event log have a newline at the end, which causes occasional blank lines in the output.

philbrookes:
> > @rhcarvalho I had intended for errors that are not part of a project (there are currently no implemented analysis of this type) would have scope of platform
>
> Does that mean we'll get confused if a project is named "platform"? 😄

Ha, good point, yes that would cause a problem alright. Maybe the solution here is scope could be either "project" or "platform" and a new field like "name" can be the project name?

philbrookes:
> @philbrookes wouldn't it be easier if we didn't have to load JSON from a file?
> One way or another, we already have the results as a Go struct/slice/idk, then we serialize it, and deserialize it back...

This seems sensible, I'll make it work this way.

output_results.go
rhcarvalho: Possibly overkill, but I knew godoc had code to rewrap and indent text:
https://play.golang.org/p/gRSnFC-gDp

rhcarvalho:
```go
package main

import (
	// ...
	"go/doc"
	// ...
)

// taken from go/src/cmd/doc/pkg.go
const (
	punchedCardWidth = 80
	indentedWidth    = punchedCardWidth - len(indent)
	indent           = "    "
)

// ...

doc.ToText(o, event.Message, "", indent, indentedWidth)
```

Though I'm not sure if that does what we want :-)

rhcarvalho:
LGTM, we can revisit the output format later if we want.

PR #49, opened 2016-09-09
Refactor analysis

rhcarvalho:
This is an attempt to simplify the code for analysis tasks, and have a single interpretation for a "task", so that we could reuse the "task runner" and have a common way to (in the near future) filter out errors that we want to log in the dump.log file, but not show to the user in stdout.

Update: this mostly reviews the type definitons used in `analysis.go`, removing the need to remap results into a map before dump, and slightly changing the format of `analysis.json`.

### DIFF
--- a/analysis.go
+++ b/analysis.go
@@ -6,28 +6,34 @@ import (
 	"path/filepath"
 )
 
-const (
-	analysisErrorNotDiscovered = iota
-	analysisErrorReadingDumpedResource
-	analysisErrorDiscoveredByAnalysis
-)
+// A CheckResult is the result of some verification of the system conditions.
+type CheckResult struct {
+	CheckName string  `json:"name"`
+	Ok        bool    `json:"ok"`
+	Message   string  `json:"message"`
+	Info      []Info  `json:"info,omitempty"`
+	Events    []Event `json:"events,omitempty"`
+}
+
+// ProjectResult stores the results of checks in a project.
+type ProjectResult struct {
+	Project string        `json:"project"`
+	Results []CheckResult `json:"checks"`
+}
+
+// AnalysisResult aggregates the result of checks executed against the system.
+// It is used to dump analysis results to a JSON file.
+type AnalysisResult struct {
+	Platform []CheckResult   `json:"platform,omitempty"`
+	Projects []ProjectResult `json:"projects,omitempty"`
+}
 
 // Info is a piece of information regarding a check, multiple Info can be
 // attached to a single Result.
 type Info struct {
-	Name      string
-	Namespace string
-	Message   string
-}
-
-// Result is a result of a single check, it can have multiple Event and Info
-// objects attached to it.
-type Result struct {
-	CheckName     string  `json:"checkName"`
-	Status        int     `json:"status"`
-	StatusMessage string  `json:"statusMessage"`
-	Info          []Info  `json:"info,omitempty"`
-	Events        []Event `json:"events,omitempty"`
+	Name      string `json:"name"`
+	Namespace string `json:"namespace"`
+	Message   string `json:"message"`
 }
 
 // Some of the types below are taken from:
@@ -97,17 +103,14 @@ type Pods struct {
 }
 
 // CheckTask is the interface which checks must implement.
-type CheckTask func(DumpedJSONResourceFactory) (Result, error)
-
-// AnalysisResults is a representation of the JSON analysis results file.
-type AnalysisResults map[string]map[string][]Result
+type CheckTask func(DumpedJSONResourceFactory) (CheckResult, error)
 
 // GetAnalysisTasks creates all the analysis tasks and sends them one by one
 // down the tasks Channel.
-func GetAnalysisTasks(tasks chan<- Task, basepath string, projects []string, results chan<- CheckResults) {
+func GetAnalysisTasks(tasks chan<- Task, basepath string, projects []string, results chan<- AnalysisResult) {
 	// Platform-wide analysis goes here.
 
-	// project specific analysis in here.
+	// Project-specific analysis goes here.
 	for _, p := range projects {
 		JSONResourceFactory := getDumpedJSONResourceFactory(filepath.Join(basepath, "projects", p))
 		tasks <- CheckProjectTask(p, results, JSONResourceFactory)
@@ -115,7 +118,7 @@ func GetAnalysisTasks(tasks chan<- Task, basepath string, projects []string, res
 }
 
 // CheckProjectTask is a task factory for tasks that diagnose system conditions.
-func CheckProjectTask(project string, results chan<- CheckResults, JSONResourceFactory DumpedJSONResourceFactory) Task {
+func CheckProjectTask(project string, results chan<- AnalysisResult, JSONResourceFactory DumpedJSONResourceFactory) Task {
 	return checkProjectTask(func() []CheckTask {
 		return []CheckTask{CheckEventLogForErrors, CheckDeployConfigsReplicasNotZero, CheckForWaitingPods}
 	}, JSONResourceFactory, project, results)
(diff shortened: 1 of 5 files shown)


analysis.go
```diff
@@ -24,83 +27,22 @@ type Info struct {
 
 // Result is a result of a single check, it can have multiple Event and Info
 // objects attached to it.
+// REVIEW: why do we have YAML annotations?
```
rhcarvalho: @philbrookes question for you, why do we have these YAML tags in the structs (Result, Events)?

analysis.go
```diff
@@ -24,83 +27,22 @@ type Info struct {
...
-		Status struct {
-			ContainerStatuses []struct {
-				Name  string `json:"name"`
-				State struct {
-					Waiting *ContainerStateWaiting `json:"waiting,omitempty"`
-				} `json:"state"`
-			} `json:"containerStatuses"`
-		} `json:"status"`
-	} `json:"items"`
+// CheckResults stores the result of a check in a scope, .... FIXME: need to review this, both Scope and Name are being set to a project name.
```
rhcarvalho: @philbrookes I tried to improve this, but couldn't get to a clear understanding of how "Scope" and "Name" are to be used. Needing help.

analysis.go
philbrookes: I was originally intending to let the analysis output as either YAML or JSON, but eventually decided that was too much effort for little value. I left these values in case we ever returned to the idea.

analysis.go
philbrookes: Scope is the scope of the check, i.e. is it checking a project or a platform (or something else in the future I suppose). The name is only really applicable to projects, it's the name of the project.

analysis.go
```diff
@@ -161,92 +100,80 @@ func checkProjectTask(checkFactory getProjectCheckFactory, JSONResourceFactory D
...
 // returned results.
-func CheckEventLogForErrors(JSONResourceFactory DumpedJSONResourceFactory) (Result, error) {
-	result := Result{Status: analysisErrorNotDiscovered, StatusMessage: "this issue was not detected", CheckName: "check eventlog for any errors", Info: []Info{}, Events: []Event{}}
-	events := Events{}
-	if err := JSONResourceFactory(filepath.Join("definitions", "events.json"), &events); err != nil {
-		result.Status = analysisErrorReadingDumpedResource
-		result.StatusMessage = "Error executing task: " + err.Error()
-		return result, err
+func CheckEventLogForErrors(events types.EventList) Result {
+	result := Result{
```
philbrookes: ah, I wondered if this was good practice or not! I wanted to do this formatting too as the line was so long, but wasn't sure it was good go, nice to know this is OK :)

tasks.go
```diff
@@ -174,6 +174,9 @@ func RunAllAnalysisTasks(runner Runner, path string, workers int) AnalysisResult
 				analysisResults["projects"][result.Name] = result.Results
 			} else if result.Scope == "platform" {
 				analysisResults["platform"]["platform"] = result.Results
+			} else {
+				// FIXME: we should handle this case properly, or code in a way that this conditional is gone.
```
rhcarvalho: @philbrookes would you have a proposal for how to handle this?
We cannot simply ignore the "else" branch, to the price of implementing new checks and having them going to limbo.

analysis.go
rhcarvalho: Depending on what you use to produce YAML, some libs look at the value for the `json` tag...
I propose getting rid of this?

analysis.go
rhcarvalho: Apparently we never dump a `CheckResults` directly, instead we convert it into a `type AnalysisResults map[string]map[string][]Result`.

analysis.go
rhcarvalho: IMHO this is much easier to work with, not just for the line length.
Note also that we're getting rid of initialization for empty stuff (Info, Events) -- instead we use the zero-value.

This is valid and idiomatic code:

```go
// foo just needs to be declared, don't need to be initialized
var foo []string
// ...
foo = append(foo, "hello")
```

https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

analysis.go
```diff
@@ -161,92 +100,80 @@ func checkProjectTask(checkFactory getProjectCheckFactory, JSONResourceFactory D
...
 		if deploymentConfig.Spec.Replicas == 0 {
-			info := Info{Name: deploymentConfig.Metadata.Name, Namespace: deploymentConfig.Metadata.Namespace, Kind: deploymentConfig.Kind, Count: 1, Message: "the replica parameter is set to 0, this should be greater than 0"}
 			result.Status = analysisErrorDiscoveredByAnalysis
 			result.StatusMessage = "one or more deployConfig replicas are set to 0"
-			result.Info = append(result.Info, info)
+			result.Info = append(result.Info, Info{
+				Name:      deploymentConfig.ObjectMeta.Name,
+				Namespace: deploymentConfig.ObjectMeta.Namespace,
+				Kind:      deploymentConfig.Kind, // REVIEW: this gets mixed information -- DeploymentConfig, Event, ..., what's the purpose?
+				Count:     1,                     // REVIEW: is Count ever going to be != 1?
```
rhcarvalho: @philbrookes one more question, what is the idea behind `Count`?
Seems that we don't increment it when we find another issue.

analysis.go
```diff
@@ -24,83 +27,22 @@ type Info struct {
...
+	Scope   string `json:"scope"`
+	Name    string `json:"name,omitempty"`
+	Results []Result
 }
 
-// CheckTask is the interface which checks must implement.
-type CheckTask func(DumpedJSONResourceFactory) (Result, error)
-
 // AnalysisResults is a representation of the JSON analysis results file.
 type AnalysisResults map[string]map[string][]Result
```
rhcarvalho: The trouble with this being a `map[string]...` is that we lose some of the type guarantees and information. One can have arbitrary string keys here, vs. a struct that gives you a prescriptive and restrictive set of keys in the output.

tasks.go
```diff
@@ -170,9 +170,9 @@ func RunAllAnalysisTasks(runner Runner, path string, workers int) AnalysisResult
 		}
 
 		for result := range checkResults {
-			if result.Scope == "project" {
-				analysisResults["projects"][result.Name] = result.Results
-			} else if result.Scope == "platform" {
+			if result.Project != "" {
```
rhcarvalho: Notice that the type definition of `CheckResults` was changed, but the JSON output has not changed.

analysis.go
rhcarvalho: Not planning to change this now, though.

analysis.go
philbrookes: This can likely be removed now, initially we were copying the count from some OpenShift output into this field, but we have removed that check since. So this field is probably useless now.

analysis.go
```diff
@@ -111,142 +52,142 @@ func GetAnalysisTasks(tasks chan<- Task, basepath string, projects []string, res
...
+
+		definition.Load("events", &events)
+		definition.Load("deploymentconfigs", &deploymentConfigs)
+		definition.Load("pods", &pods)
+
+		if err := definition.Err(); err != nil {
+			// REVIEW: do we want / is there a clean way to have
+			// analysisErrorReadingDumpedResource status in the
+			// output? Alternatively, returning an error will get it
+			// into dump.log and stdout.
```
rhcarvalho: @philbrookes we need to review the fate of `analysisErrorReadingDumpedResource`. Now that we load resources independent of running a `Check*`, there is no result where to put that error.

I think I need to further change this, because with the current implementation the task aborts on error loading any of the resources, but we want to attempt running each check independently.

analysis.go
rhcarvalho: Removed!

analysis.go
```diff
@@ -111,142 +51,140 @@ func GetAnalysisTasks(tasks chan<- Task, basepath string, projects []string, res
...
 	for _, deploymentConfig := range deploymentConfigs.Items {
 		if deploymentConfig.Spec.Replicas == 0 {
-			info := Info{Name: deploymentConfig.Metadata.Name, Namespace: deploymentConfig.Metadata.Namespace, Kind: deploymentConfig.Kind, Count: 1, Message: "the replica parameter is set to 0, this should be greater than 0"}
 			result.Status = analysisErrorDiscoveredByAnalysis
 			result.StatusMessage = "one or more deployConfig replicas are set to 0"
-			result.Info = append(result.Info, info)
+			result.Info = append(result.Info, Info{
+				Name:      deploymentConfig.ObjectMeta.Name,
+				Namespace: deploymentConfig.ObjectMeta.Namespace,
+				Kind:      deploymentConfig.Kind, // REVIEW: this gets mixed information -- DeploymentConfig, Event, ..., what's the purpose?
```
rhcarvalho: @philbrookes, item for review -- what's the intent behind `Info.Kind`? I see here it gets populated via `deploymentConfig.Kind`, that given valid JSON output from `oc` will eventually be "DeploymentConfig".

Elsewhere we hard code `Info.Kind` to "container".
https://github.com/feedhenry/fh-system-dump-tool/pull/49/files#diff-93027c7b7b59812efa3717a7ef539251R140

philbrookes: This was added (similar to count), before we had any real grasp on what the analysis side of this tool was going to look like. I believe the idea here was to relate to the user what type of OpenShift object the info was related to i.e. pods or pvs or whatever. In hindsight it seems to be an unrequired field though.

rhcarvalho: Okay, thanks. I've removed it as well.

philbrookes:
This LGTM, ran locally and all analysis functions as usual.

PR #50, opened 2016-09-09
Do not show some errors in stdout

rhcarvalho:
We want a way to log errors to dump.log, while not bothering users with
common errors that do not reflect actual problems.

For instance, it is common that a call to fetch previous logs will fail
to find a previous pod.

Another example are commands that are available in OSE 3.3 but not in
OSE 3.2, and those are executed regardless of the current client version
in a best effort manner.

---

https://issues.jboss.org/browse/RHMAP-9831

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -18,3 +18,40 @@ func (e errorList) Error() string {
 	}
 	return "multiple errors:\n" + strings.Join(msgs, "\n")
 }
+
+// IgnorableError is an error that has an extra method to tell whether it should
+// be ignored. Ignored errors may be omitted from standard program output.
+type IgnorableError interface {
+	error
+	Ignore() bool // Is the error a timeout?
+}
+
+// ignoredError implements IgnorableError, always returning true.
+type ignoredError struct {
+	err error
+}
+
+func (e *ignoredError) Error() string {
+	if e.err == nil {
+		return ""
+	}
+	return e.err.Error()
+}
+
+// Ignore implements IgnorableError.
+func (e *ignoredError) Ignore() bool {
+	return true
+}
+
+// Assert that ignoredError implements the IgnorableError interface.
+var _ IgnorableError = (*ignoredError)(nil)
+
+// MarkErrorAsIgnorable marks the original error as ignored if non-nil.
+func MarkErrorAsIgnorable(err error) error {
+	if err == nil {
+		// Keeps nil errors nil, to avoid the confusion of having a
+		// non-nil type and a nil value in an interface value.
+		return nil
+	}
+	return &ignoredError{err}
+}
--- a/logs.go
+++ b/logs.go
@@ -121,6 +121,6 @@ func ocLogs(r Runner, resource LoggableResource, maxLines int, extraArgs []strin
 			filename += "_" + resource.Container
 		}
 		path := filepath.Join("projects", resource.Project, what, filename+".logs")
-		return r.Run(cmd, path)
+		return MarkErrorAsIgnorable(r.Run(cmd, path))
 	}
 }
--- a/main.go
+++ b/main.go
@@ -4,6 +4,7 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -81,6 +82,18 @@ func archive(path string) error {
 	return cmd.Run()
 }
 
+func checkPrerequisites() error {
+	if _, err := exec.LookPath("oc"); err != nil {
+		return errors.New("oc command not found, please install the OpenShift CLI before using this tool")
+	}
+	// This serves as a sort of "ping" to the server, to make sure we are
+	// logged in and can access the server before we issue more calls.
+	if err := exec.Command("oc", "whoami").Run(); err != nil {
+		return errors.New("could not access OpenShift, please run 'oc login' and make sure that a user is logged in and the server is accessible")
+	}
+	return nil
+}
+
 func main() {
 	flag.Parse()
 
@@ -92,6 +105,10 @@ func main() {
 	if !(*concurrentTasks > 0) {
 		fmt.Fprintln(os.Stderr, "Error: argument to -p flag must be greater than 0")
 		os.Exit(1)
+	}
+
+	if err := checkPrerequisites(); err != nil {
+		log.Fatalln("Error:", err)
 	}
 
 	start := time.Now().UTC()
--- a/main.go
+++ b/main.go
@@ -142,7 +142,7 @@ func main() {
 	runner := NewDumpRunner(basePath)
 
 	log.Print("Running dump and analyse tasks...")
-	RunAllDumpTasks(runner, basePath, *concurrentTasks)
+	RunAllDumpTasks(runner, basePath, *concurrentTasks, fileOnlyLogger)
 	analysisResults := RunAllAnalysisTasks(runner, basePath, *concurrentTasks)
 
 	delta := time.Since(start)
--- a/tasks.go
+++ b/tasks.go
@@ -14,9 +14,14 @@ import (
 // A Task performs some part of the RHMAP System Dump Tool.
 type Task func() error
 
+// Logger is a minimal interface implemented by log.Logger.
+type Logger interface {
+	Printf(format string, v ...interface{})
+}
+
 // RunAllDumpTasks runs all tasks known to the dump tool using concurrent workers.
 // Dump output goes to path.
-func RunAllDumpTasks(runner Runner, path string, workers int) {
+func RunAllDumpTasks(runner Runner, path string, workers int, fileOnlyLogger Logger) {
 	tasks := GetAllDumpTasks(runner, path)
 	results := make(chan error)
 
@@ -41,6 +46,10 @@ func RunAllDumpTasks(runner Runner, path string, workers int) {
 	// Loop through the task execution results and log errors.
 	for err := range results {
 		if err != nil {
+			if ierr, ok := err.(IgnorableError); ok && ierr.Ignore() {
+				fileOnlyLogger.Printf("Task error: %v", err)
+				continue
+			}
 			// TODO: there should be a way to identify which task
 			// had an error.
 			fmt.Fprintln(os.Stderr)
--- a/definitions.go
+++ b/definitions.go
@@ -39,6 +39,6 @@ func ResourceDefinition(r Runner, project, resource string) Task {
 		tree = append(tree, "definitions", fname)
 
 		path := filepath.Join(tree...)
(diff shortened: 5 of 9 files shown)


tasks.go
```diff
@@ -14,9 +14,14 @@ import (
...
 
+// Logger is a minimal interface implemented by log.Logger.
+type Logger interface {
+	Printf(format string, v ...interface{})
+}
+
 // RunAllDumpTasks runs all tasks known to the dump tool using concurrent workers.
 // Dump output goes to path.
-func RunAllDumpTasks(runner Runner, path string, workers int) {
+func RunAllDumpTasks(runner Runner, path string, workers int, fileOnlyLogger Logger) {
```
rhcarvalho: I would like to introduce a "TaskRunner", it could then take a `logger` and a `fileOnlyLogger` as attributes.
Though for that would be nice to unify the code in this file first.

errors.go
```diff
@@ -18,3 +18,38 @@ func (e errorList) Error() string {
...
+		return ""
+	}
+	return e.err.Error()
+}
+
+func (e *ignoredError) Ignore() bool {
+	return true
+}
+
+var _ IgnorableError = (*ignoredError)(nil)
```
philbrookes: What is the purpose of this line of code?

rhcarvalho: To ensure in compile time that `*ignoredError` implements the `IgnorableError` interface.

logs.go
```diff
@@ -91,18 +91,23 @@ func GetPodContainers(runner Runner, project, name string) ([]string, error) {
...
 }
 
 // ocLogs fetches logs from an OpenShift resource using oc.
-func ocLogs(r Runner, resource LoggableResource, maxLines int, extraArgs []string, what string) Task {
-	return func() error {
+func ocLogs(r Runner, resource LoggableResource, maxLines int, extraArgs []string, what string, ignoreErrors bool) Task {
+	return func() (err error) {
+		defer func() {
+			if ignoreErrors {
+				err = IgnoreError(err)
```
philbrookes: This deferred function is casting the returned error to an `IgnoreError` right before the function exits? That is really cool! I did not know this could be done.

errors.go
rhcarvalho: Example: https://github.com/golang/go/blob/master/src/crypto/aes/aes_gcm.go#L51-L52

errors.go
```diff
@@ -18,3 +18,38 @@ func (e errorList) Error() string {
...
+}
+
+func (e *ignoredError) Error() string {
+	if e.err == nil {
+		return ""
+	}
+	return e.err.Error()
+}
+
+func (e *ignoredError) Ignore() bool {
```
philbrookes: Does `Ignored` make more sense here?
```
if error.Ignored() {
   // ignore the error
}
```

errors.go
philbrookes: Ah I see, that's nifty

rhcarvalho: The stdlib has even more examples:

https://github.com/golang/go/blob/master/src/encoding/json/stream.go#L263-L264
https://github.com/golang/go/blob/master/src/io/multi.go#L64
https://github.com/golang/go/blob/master/src/net/http/server.go#L1443
https://github.com/golang/go/blob/master/src/testing/testing.go#L382-L383

```
$ grep -Er 'var _ .*\(nil\)$' `go env GOROOT`/src
/usr/local/go/src/io/multi.go:var _ stringWriter = (*multiWriter)(nil)
/usr/local/go/src/encoding/json/stream.go:var _ Marshaler = (*RawMessage)(nil)
/usr/local/go/src/encoding/json/stream.go:var _ Unmarshaler = (*RawMessage)(nil)
/usr/local/go/src/encoding/json/decode_test.go:var _ encoding.TextUnmarshaler = (*unmarshalerText)(nil)
/usr/local/go/src/encoding/json/decode_test.go:var _ encoding.TextUnmarshaler = (*u8marshal)(nil)
/usr/local/go/src/encoding/xml/marshal_test.go:var _ Marshaler = (*MyMarshalerTest)(nil)
/usr/local/go/src/encoding/xml/marshal_test.go:var _ MarshalerAttr = (*MyMarshalerAttrTest)(nil)
/usr/local/go/src/encoding/xml/read_test.go:var _ Unmarshaler = (*MyCharData)(nil)
/usr/local/go/src/encoding/xml/read_test.go:var _ UnmarshalerAttr = (*MyAttr)(nil)
/usr/local/go/src/net/http/server.go:var _ closeWriter = (*net.TCPConn)(nil)
/usr/local/go/src/testing/testing.go:var _ TB = (*T)(nil)
/usr/local/go/src/testing/testing.go:var _ TB = (*B)(nil)
/usr/local/go/src/crypto/aes/ctr_s390x.go:var _ ctrAble = (*aesCipherAsm)(nil)
/usr/local/go/src/crypto/aes/cbc_s390x.go:var _ cbcEncAble = (*aesCipherAsm)(nil)
/usr/local/go/src/crypto/aes/cbc_s390x.go:var _ cbcDecAble = (*aesCipherAsm)(nil)
/usr/local/go/src/crypto/aes/aes_gcm.go:var _ gcmAble = (*aesCipherGCM)(nil)
```

:smile:

philbrookes:
:+1: LGTM.

errors.go
rhcarvalho: Yeah, I was pondering it. My line of thought was that the error is only ignored after you ignore it :)
IOW `Ignore()` returns whether I may ignore the error or not.

I started sketching this out from the interface definition in https://github.com/feedhenry/fh-system-dump-tool/pull/50/files#diff-dd25e0e0a594c3f3592c1c717039b85eR22. There I say what the method is about.
If we make it `Ignored()`, wouldn't it sound like an error has already been ignored (and now it is "too late to make a decision")?

logs.go
rhcarvalho: Yeah, it wraps the original error with another value.

Deferred calls can mutable named return values. Not something we want to abuse though, as it gets more confusing to reason about.

Just went with it here but would also work to change the last line to add an if there.
The problem is that if you have multiple return points, then you need to remember to add that logic in every one of them -- or use the deferred call.

errors.go
philbrookes: I guess both words are a bit confusing, I feel like `error.Ignore()` is instructing the error to be ignore-able (like I'd imagine `error.Ignore(true)`). How about `CanIgnore`?

philbrookes:
Local output with 3.2:
```
16:36 $ ./fh-system-dump-tool
2016/09/09 16:36:29 Starting RHMAP System Dump Tool...
2016/09/09 16:36:29 Running dump and analyse tasks...

2016/09/09 16:36:30 Task error: command "oc policy can-i list persistentvolumes": exit status 1: error: unknown command "can-i list persistentvolumes"
See 'oc policy -h' for help and examples.

2016/09/09 16:36:30 Task error: command "oc policy can-i list nodes": exit status 1: error: unknown command "can-i list nodes"
See 'oc policy -h' for help and examples.
.
2016/09/09 16:36:30 Task error: command "oc policy can-i list projects": exit status 1: error: unknown command "can-i list projects"
See 'oc policy -h' for help and examples.
.................................................
2016/09/09 16:36:35 Task error: command "oc adm diagnostics": exit status 255:
.........................................................
2016/09/09 16:36:41 Task error: command "oc get persistentvolumes -o=json": exit status 1: Error from server: User "admin" cannot list all persistentvolumes in the cluster
.
2016/09/09 16:36:41 Task error: command "oc get nodes -o=json": exit status 1: Error from server: User "admin" cannot list all nodes in the cluster
..................................................
2016/09/09 16:36:57 Task error: command "oc -n logging logs pods/logging-deployer-deaal -c deployer --tail 1000": exit status 1: Error from server: API error (404): no such id: d37a4b808ae1290a5cd12d4f5dd18e1812c973bb1dbd2d6b5f13b98e524b6079
................................................
...............
2016/09/09 16:37:37 Run all tasks in 1m7s.
```

rhcarvalho:
Omitting more of the cluster-level errors now (oc policy, version, whoami).

philbrookes:
output is much tidier now:
```
16:52 $ ./fh-system-dump-tool
2016/09/09 16:52:39 Starting RHMAP System Dump Tool...
2016/09/09 16:52:39 Running dump and analyse tasks...
..................................................................
2016/09/09 16:52:45 Task error: command "oc adm diagnostics": exit status 255:
...........................
2016/09/09 16:52:48 Task error: command "oc get nodes -o=json": exit status 1: Error from server: User "admin" cannot list all nodes in the cluster

2016/09/09 16:52:48 Task error: command "oc get persistentvolumes -o=json": exit status 1: Error from server: User "admin" cannot list all persistentvolumes in the cluster
......................................
2016/09/09 16:53:00 Task error: command "oc -n logging logs pods/logging-deployer-deaal -c deployer --tail 1000": exit status 1: Error from server: API error (404): no such id: d37a4b808ae1290a5cd12d4f5dd18e1812c973bb1dbd2d6b5f13b98e524b6079
......................................................................
...............
2016/09/09 16:53:41 Run all tasks in 1m1s.
```

rhcarvalho:
Added some extra documentation comments around the discussion about the `IgnorableError` interface.
Augmenting the error interface mimics this idea from the Go stdlib: https://golang.org/pkg/net/#Error

rhcarvalho:
> command "oc -n logging logs pods/logging-deployer-deaal -c deployer --tail 1000": exit status 1: Error from server: API error (404): no such id: ...

@philbrookes I wonder if we are in a point or state of mind to omit from stdout all errors related to fetching logs?

rhcarvalho:
After discussion on IRC we decided to proceed with omitting from stderr most things, except:

- 'oc' binary is missing (can't really access the cluster to fetch data)
- user is not logged in with oc (again, can't fetch data)
- 'tar' binary is missing (cannot generate final archive)
- a Nagios pod is missing (discussed previously, we've decided to explicitly warn users that they should have Nagios running to provide us more useful data about the system)

The first three will be checked as a prerequisite to running the tool.

main.go
```diff
@@ -81,6 +82,21 @@ func archive(path string) error {
 	return cmd.Run()
 }
 
+func checkPrerequisites() error {
+	if _, err := exec.LookPath("oc"); err != nil {
```
rhcarvalho: We could omit this deferring to only call `exec.Command("oc", "whoami").Run()`, but then we'd have to have a longer, more generic error message.

rhcarvalho:
@philbrookes, I've marked a lot of potential errors as ignorable, plus added precodition checking (Do we have oc? Can we access the server?).

PTAL

philbrookes:
LGTM :+1:

PR #51, opened 2016-09-12
go vet: usage of t.Fatalf with non-constant format

rhcarvalho:
This fixes a warning from go vet. Normally we pass a constant string
literal to t.Fatalf.

In this case we can also compare errors directly, instead of comparing
their string representation.


philbrookes:
:+1:

PR #52, opened 2016-09-12
Rewrite analysis tasks

rhcarvalho:
Simplify code and tests moving away from factories and mocks.

This makes the analysis checks take structured data, becoming easy to test and implement.

Take API type definitions from OpenShift/Kubernetes, moving away from embedded anonymous structs which were notably hard to instantiate.

### DIFF
--- a/analysis.go
+++ b/analysis.go
@@ -2,17 +2,20 @@ package main
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"path/filepath"
+
+	"github.com/feedhenry/fh-system-dump-tool/openshift/api/types"
 )
 
 // A CheckResult is the result of some verification of the system conditions.
 type CheckResult struct {
-	CheckName string  `json:"name"`
-	Ok        bool    `json:"ok"`
-	Message   string  `json:"message"`
-	Info      []Info  `json:"info,omitempty"`
-	Events    []Event `json:"events,omitempty"`
+	CheckName string        `json:"name"`
+	Ok        bool          `json:"ok"`
+	Message   string        `json:"message"`
+	Info      []Info        `json:"info,omitempty"`
+	Events    []types.Event `json:"events,omitempty"`
 }
 
 // ProjectResult stores the results of checks in a project.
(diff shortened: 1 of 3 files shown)


philbrookes:
@rhcarvalho when I run this branch I still get all the error output, shouldn't stderr be much quieter now?

analysis.go
```diff
@@ -36,222 +39,141 @@ type Info struct {
...
+type definitionLoader struct {
+	basepath, project string
+	err               error
+}
+
+func (l *definitionLoader) Load(kind string, v interface{}) {
+	if l.err != nil {
+		return
+	}
+	path := filepath.Join(l.basepath, "projects", l.project, "definitions", kind+".json")
```
philbrookes: Is there a reason this only loads definitions, and not logs, or logs-previous, or nagios, etc?

Could it instead be a `projectEntityLoader` with `LoadDefinition`, `loadLog`, etc methods?

rhcarvalho:
> @rhcarvalho when I run this branch I still get all the error output, shouldn't stderr be much quieter now?

That's because it is not rebase on top of latest master, with the log changes in.

If you pull `refs/pull/52/merge`, it should be quieter.

I can rebase and push force as well if you prefer.

philbrookes:
> That's because it is not rebase on top of latest master, with the log changes in.

Does it make sense to do that before merging into master?

rhcarvalho:
> > That's because it is not rebase on top of latest master, with the log changes in.
> 
> Does it make sense to do that before merging into master?

Travis runs tests against the merged branch, not the PR head:

https://travis-ci.org/feedhenry/fh-system-dump-tool/builds/159418417#L120

That's what I would expect from CI anyway :)

We only need a manual rebase if there are conflicts to be resolved. Otherwise every contributor would need to be constantly rebasing while waiting in the merge queue of a busy repository...

analysis.go
rhcarvalho: One advantage of being a `definitionLoader` is that we can talk in terms of the `DefinitionLoader` interface, which could really be called `Loader` if we were to have specific loaders for other things.

What bothers me more here is that we have the knowledge about where definitions are stored in two different places: here and where they are written (by the DumpRunner).

philbrookes:
> Travis runs tests against the merged branch, not the PR head

Awesome!

philbrookes:
:+1: LGTM

PR #54, opened 2016-09-13
Improve CI, allow running locally

rhcarvalho:
- Use goimports in addition to gofmt -s.
- Checks `go vet` for errors.
- Experimentally adds `golint` -- we've discussed adding it before, we
  can give it a shot and disable if it starts to annoy us.
- Run tests an additional time with the -race flag.
- Document how to test code locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+SHELL := /bin/bash
+
 ifndef FH_SYSTEM_DUMP_TOOL_VERSION
 FH_SYSTEM_DUMP_TOOL_VERSION := $(shell git describe --tags --abbrev=14)
 endif
@@ -10,3 +12,32 @@ all:
 .PHONY: clean
 clean:
 	@-go clean -i
+
+.PHONY: ci
+ci: check-gofmt check-goimports check-golint vet test test-race
+
+# goimports doesn't support the -s flag to simplify code, therefore we use both
+# goimports and gofmt -s.
+.PHONY: check-gofmt
+check-gofmt:
+	diff <(gofmt -s -d .) <(printf "")
+
+.PHONY: check-goimports
+check-goimports:
+	diff <(goimports -d .) <(printf "")
+
+.PHONY: check-golint
+check-golint:
+	diff <(golint ./...) <(printf "")
+
+.PHONY: vet
+vet:
+	go vet ./...
+
+.PHONY: test
+test:
+	go test -v -cpu=2 ./...
+
+.PHONY: test-race
+test-race:
+	go test -v -cpu=1,2,4 -short -race ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: go
 go:
- - 1.6.3
+  - 1.6.3
+install:
+  - go get -u github.com/golang/lint/golint
+  - go get -u golang.org/x/tools/cmd/goimports
 script:
-  - test -z "$(gofmt -s -l $(find . -name '*.go'))" || { echo "Code needs to be gofmt'ed"; false; }
-  - go test ./...
+  - make --keep-going ci
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,6 +29,24 @@ directory. It is recommended that GOPATH/bin is part of your PATH environment
 variable.
 
 
+## Testing
+
+Pull Requests are automatically tested using [Travis
+CI](https://travis-ci.org/). You can run the same tests locally during development:
+
+```
+make -k ci
+```
+
+That will run several verifications involving code formatting, unit tests, etc.
+The `-k` flag tells `make` to keep running even if one of the verifications
+fail. If you want to to terminate after encountering the first problem, omit
+that flag.
+
+To see what commands are used for testing, look at the output of `make -n ci`.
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -10,3 +10,32 @@ all:
...
 	@-go clean -i
+
+.PHONY: ci
+ci: check-gofmt check-goimports check-golint vet test test-race
+
+# goimports doesn't support the -s flag to simplify code, therefore we use both
+# goimports and gofmt -s.
+.PHONY: check-gofmt
+check-gofmt:
+	bash -c 'diff <(gofmt -s -d .) <(printf "")'
```
rhcarvalho: @david-martin check this out, just learned a nicer way to make sure `gofmt` etc fail if the output is not empty :)

Makefile
```diff
@@ -10,3 +10,32 @@ all:
...
+
+.PHONY: check-goimports
+check-goimports:
+	bash -c 'diff <(goimports -d .) <(printf "")'
+
+.PHONY: check-golint
+check-golint:
+	bash -c 'diff <(golint ./...) <(printf "")'
```
philbrookes: Wohoo :) Hope this works out!

Makefile
rhcarvalho: Setting `SHELL := /bin/bash` we can get rid of `bash -c`... will update.

Makefile
rhcarvalho: Yeah, so far it does :)

I learned the diff trick from the internets... tested it, works as expected (to detect actual problems).

Makefile
david-martin: @rhcarvalho SORCERY! :)

I had to understand.
```
$ diff <(echo "x") <(echo "y")
1c1
< x
---
> y
$ echo $?
1

$ diff <(echo "x") <(echo "x")
$ echo $?
0
```

I like it :+1:

rhcarvalho:
Introducing mistakes in the Go files shows how each of those checks is useful.

<details>
```console
$ git diff -U1
diff --git a/definitions_test.go b/definitions_test.go
index 6600eee..86743b4 100644
--- a/definitions_test.go
+++ b/definitions_test.go
@@ -45,3 +45,3 @@ func TestResourceDefinition(t *testing.T) {
        }
-       for i, tt := range tests {
+       for i, tt := range tests[:len(tests)] {
                runner := &FakeRunner{}
diff --git a/dump_writers.go b/dump_writers.go
index 798390d..afe5a63 100644
--- a/dump_writers.go
+++ b/dump_writers.go
@@ -27,2 +27,4 @@ func outToFile(basepath, extension, scope string) projectResourceWriterCloserFac
                return f, f, nil
+               if true {
+               }
        }
diff --git a/logs.go b/logs.go
index 631a332..7609ef2 100644
--- a/logs.go
+++ b/logs.go
@@ -9,2 +9,7 @@ import (
 
+// Check for error in the wrong location, error should be the last type.
+func dummyfunc() (error, int) {
+       return nil, 0
+}
+
 // LoggableResource describes an OpenShift resource that produces logs. Even
diff --git a/main.go b/main.go
index 0e50988..4a26097 100644
--- a/main.go
+++ b/main.go
@@ -9,2 +9,3 @@ import (
        "fmt"
+       _ "github.com/dummy/import"
        "io"
```
</details>

```console
$ # run just a subset of `make -k ci`
$ make -k check-gofmt check-goimports check-golint vet
diff <(gofmt -s -d .) <(printf "")
1,12d0
< diff definitions_test.go gofmt/definitions_test.go
< --- /tmp/gofmt335090355       2016-09-13 17:28:09.290100982 +0200
< +++ /tmp/gofmt193666166       2016-09-13 17:28:09.290100982 +0200
< @@ -43,7 +43,7 @@
<                       },
<               },
<       }
< -     for i, tt := range tests[:len(tests)] {
< +     for i, tt := range tests[:] {
<               runner := &FakeRunner{}
<               task := ResourceDefinition(runner, tt.project, tt.resource)
<               if err := task(); err != nil {
Makefile:23: recipe for target 'check-gofmt' failed
make: *** [check-gofmt] Error 1
diff <(goimports -d .) <(printf "")
1,20d0
< diff main.go gofmt/main.go
< --- /tmp/gofmt267049300       2016-09-13 17:28:09.338100011 +0200
< +++ /tmp/gofmt357755043       2016-09-13 17:28:09.338100011 +0200
< @@ -7,7 +7,6 @@
<       "errors"
<       "flag"
<       "fmt"
< -     _ "github.com/dummy/import"
<       "io"
<       "io/ioutil"
<       "log"
< @@ -16,6 +15,8 @@
<       "path/filepath"
<       "runtime"
<       "time"
< +
< +     _ "github.com/dummy/import"
<  )
<  
<  const (
Makefile:27: recipe for target 'check-goimports' failed
make: *** [check-goimports] Error 1
diff <(golint ./...) <(printf "")
1d0
< logs.go:11:1: error should be the last type when returning multiple items
Makefile:31: recipe for target 'check-golint' failed
make: *** [check-golint] Error 1
go vet ./...
dump_writers.go:28: unreachable code
exit status 1
Makefile:35: recipe for target 'vet' failed
make: *** [vet] Error 1
```

Yes, quite verbose. At least it gives us the information we care about.

philbrookes:
:+1: LGTM

PR #56, opened 2016-09-16
Do not print error from GetPodContainers to user

rhcarvalho:
Log the error to the dump.log file, but do not bother the user, as she
can't act upon that error.

Completes the work from https://github.com/feedhenry/fh-system-dump-tool/pull/50.


philbrookes:
:+1:

PR #57, opened 2016-09-16
Fine tune output messages

rhcarvalho:
- Rephrase messages trying to be more user friendly
- Print errors at the end to avoid interleaving with dots "."
- Print an "all is good" message if the analysis found no problems
- Add a way to force the output of all errors to stderr (disable the "omit error" feature; intended for development-only)


main.go
```diff
@@ -158,14 +163,28 @@ func main() {
 
 	runner := NewDumpRunner(basePath)
 
-	log.Print("Running dump and analyse tasks...")
-	RunAllDumpTasks(runner, basePath, *concurrentTasks, fileOnlyLogger)
+	log.Print("Collecting system information...")
```
philbrookes: This is much better phrasing

philbrookes:
:+1: LGTM

PR #58, opened 2016-09-19
Add make target to build a release

rhcarvalho:
This is what I used to generate https://github.com/feedhenry/fh-system-dump-tool/releases/tag/v1.0.0.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,9 @@ FH_SYSTEM_DUMP_TOOL_VERSION := $(shell git describe --tags --abbrev=14)
 endif
 LDFLAGS := -X main.Version=$(FH_SYSTEM_DUMP_TOOL_VERSION)
 
+IMPORT_PATH := github.com/feedhenry/fh-system-dump-tool
+GO_DOCKER_IMAGE := golang:1.8
+
 .PHONY: all
 all:
 	@go install -v -ldflags '$(LDFLAGS)'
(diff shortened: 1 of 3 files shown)


rhcarvalho:
@philbrookes, though this is approaching its first birthday, it might still be useful. Shall we merge it in?

philbrookes:
@rhcarvalho This looks great, I don't know it was overlooked for so long!

Makefile
```diff
@@ -5,6 +5,9 @@ FH_SYSTEM_DUMP_TOOL_VERSION := $(shell git describe --tags --abbrev=14)
 endif
 LDFLAGS := -X main.Version=$(FH_SYSTEM_DUMP_TOOL_VERSION)
 
+IMPORT_PATH := github.com/feedhenry/fh-system-dump-tool
+GO_DOCKER_IMAGE := golang:1.6
```
rhcarvalho: This is [not available anymore](https://hub.docker.com/_/golang/). Need to test, but it should work with any newer version of Go as well.

philbrookes: If this works with 1.8, is there any reason not to update to that?

rhcarvalho: Tested and updated, it works just fine.

rhcarvalho:
@philbrookes thanks for having a look. If I have your approval we can finally merge ;)

philbrookes:
@rhcarvalho :+1:

